PR #251, opened 2024-07-04
[FEATURE] Separate steel cuffs from the spreader bar

SandrinePDR:
This PR will separate the steel cuffs from the spreader bar item and adds the bar and the chains as separate modules to the steel cuffs.
See: #249


SandrinePDR:
Closed due to strange GitHub behavior

PR #271, opened 2024-08-30
[ADD] New asset steel wrist cuffs

SandrinePDR:
## References

_None_

## About The Pull Request

Added a new steel wrist cuffs asset that works the same as the leather ones, including the belt chain options

## Changelog

Authored by: Sandrine & Claudia (original cuff graphics)

```md
Assets:
- [Steel Wrist Cuffs] New asset
```

## Checklist


ClaudiaMia (approved):
Looking and working great!

I just changed the preview image to be compliant with the intended but not fully clear and complete instructions.

The goal is to have a transparent grey-ish stylized body that has no resemblance to a body that might violate the GitHub terms of use.

I also fixed a bug with the chained overhead pose that was also already present on the leather cuffs, where I fixed it, too.

And during my testing I found the same kind of not reaching chains issue on the bondage frame, so I fixed that, too, while I was at it.

Please tell me if you are fine with me doing these changes here and then I will merge.

SandrinePDR:
I'm fine in general (and shouldn't complain as you are the original creator of the asset's base), but it could be worth it to change the steel cuff's ring positions in the over head pose. Then the chains would have an anchor point. But we/ I can also do this at a later point in time.
So please merge the PR, if your good.

PR #273, opened 2024-08-31
[ADD] New asset "Connecting Chains"

SandrinePDR:
## References

_None_

## About The Pull Request

* This PR moves the standalone "Belt Chains" into a new more flexible asset called "Connecting Chains". 
* The belt chains are still available, but now as a module instead of a standalone asset. 
* Second module are "Armbinder Chains" that connect the ring of an armbinder to worn ankle cuffs, forcing the player into a kneeling position and limiting their ability to spread their legs

image

Drawback is that with the release of the asset, existing belt chains will probably vanish into thin air.
There are two known graphical glitches for leg spread poses, but that is minor compared to the allowed flexibility in my opinion.

## Changelog

Authored by: Sandrine

```md
Assets:
- [Connecting Chains] A new way to connect the items of your "victims". 
 * Chain their naughty hands to their chastity blets to keep them out of the way 
 * Thread a chain through the Ring of an armbinder and force them to stay on their knees.
```

## Checklist


src/assets/misc/connectingChains/connectingChains.asset.ts
```diff
@@ -0,0 +1,107 @@
...
+								'Ankle_cuffs_chainable',
+								'Armbinder_chainable',
+							],
+						},
+					},
+				},
+			],
+		},
+	},
+	chat: {
```
ClaudiaMia, line 89: These messages are no longer correct. 

You could also remove the 'chat' property completely, then default neutral asset add/remove messages are used, plus your switch messages from the module changes.

src/assets/misc/connectingChains/connectingChains.asset.ts
```diff
@@ -0,0 +1,107 @@
...
+				{
+					id: 'armbinderChain',
+					name: 'Kneeling Chains',
+					switchMessage: 'SOURCE_CHARACTER connects the cuffs to TARGET_CHARACTER_DYNAMIC_POSSESSIVE armbinder with short chains.',
+					properties: {
+						poseLimits: {
+							legs: 'kneeling',
+							bones:
+								{
+									leg_l: [[-28, 5]],
```
ClaudiaMia, line 70: You should probably decrease the min value to -26 or -27, as at -28 the chains disappear.

src/assets/misc/connectingChains/connectingChains.asset.ts
```diff
@@ -0,0 +1,107 @@
...
+							],
+						},
+					},
+				},
+				{
+					id: 'armbinderChain',
+					name: 'Kneeling Chains',
+					switchMessage: 'SOURCE_CHARACTER connects the cuffs to TARGET_CHARACTER_DYNAMIC_POSSESSIVE armbinder with short chains.',
+					properties: {
+						poseLimits: {
```
ClaudiaMia, line 66: Please test the front view, too.
In the front view one can still see the chains and it looks clearly wrong.

The best way is to hide the chain images when the view is front and add new chain images for just the front view.
I do not think that you need many new chain variants to show only in the front facing view. Maybe one or two should be enough so that they disappear behind the legs, but do not clip through on the other side when manually dragging the slider from 5 to -27.

ClaudiaMia (changes requested):
Aside from my finds, this is looking and working really well!

I feel in the future, we could add more variants to this asset, for instance one that chains ankle and wrist cuffs together behind the back while kneeling.

I really like this asset!

src/assets/misc/connectingChains/connectingChains.asset.ts
SandrinePDR: Good find. I really haven't thought about this.

ClaudiaMia (approved):
Great work!

PR #276, opened 2024-09-03
[ADD] New Asset: Leggings

SandrinePDR:
## References

_None_

## About The Pull Request

Adds a pair of leggings as a new asset
image

## Changelog

Authored by: Claudia (base design) & Sandrine (leggings)

```md
Assets:
- [Leggings] New
```

## Checklist


src/assets/legwear/leggings/leggings.asset.ts
```diff
@@ -0,0 +1,51 @@
...
+			name: 'Leggings',
+			default: '#2A2A2A',
+		},
+	},
+	// size:800, y:574, centered
+	preview: 'preview.png',
+	attributes: {
+		provides: [
+			'Clothing',
+			'Legwear',
```
ClaudiaMia, line 16: I guess this is a matter definition and not really about right/wrong. Just making sure you did it intentionally.

I could not find the leggings, as I was looking in the trousers, pants, and skirts section which is `Clothing_lower`, not in the `Legwear section`, which means socks, stockings, and pantyhoses. 

The intended difference was things worn mainly over just the legs, but potentially going over the hops, but primarily "underwear", whereas everything that can be worn over the hips and down the legs but is a primary clothing on the outside, is the other section.

src/assets/legwear/leggings/leggings.asset.ts
```diff
@@ -0,0 +1,51 @@
...
+			type: 'typed',
+			name: 'Leggings design',
+			variants: [
+				{
+					id: 'netshowing',
+					name: 'Net showing',
+					default: true,
+				},
+				{
+					id: 'plain',
```
ClaudiaMia, line 30: Another thing I noticed, where I was not sure if intended or not:

The plain setting only covers the gap on the front, but not the small one on the back.

src/assets/legwear/leggings/leggings.asset.ts
SandrinePDR: Moved it to the bottom category for consistancy

src/assets/legwear/leggings/leggings.asset.ts
SandrinePDR: Good catch. Somehow the 'fill' image got broken during export.

ClaudiaMia (approved):
LGTM!

PR #293, opened 2024-09-27
[ADD] Generic poster asset

SandrinePDR:
## References

_None_ (a lot of in game references and events)

## About The Pull Request

Added a fully customizable advertisement of Towel & Towel ™ , the club's main sponsor and supplier of towels as a proof of concept for the new generic poster asset

## Changelog

Authored by: Taja, Claudia & Sandrine

```md
Assets:
- [Poster] A new asset to decorate your room with
```

## Checklist


ClaudiaMia (changes requested):
I talked with Clare and we suggest to change the name of the asset to something neutral like "Poster" and then adding a module with just one option which is the towel and towel advertisement so that it is clear for other asset creators that future similar things are to be added to the module of this asset.

The color names should then also be a neutral sounding background 1, background 2, color 1, color 2.... Something along these lines~

Does that make sense for you, is it fine?

One last minor thing that I noticed and where I assume it is not intentional: The image is not centered inside the passe-partout, or rather, that frame is thicker on the right side.

ClaudiaMia (approved):
LGTM!

PR #294, opened 2024-09-27
[ADD] Set of steel spheres for hands and head

SandrinePDR:
## References

_None_

## About The Pull Request

Added two steel spheres as mittens and a wider sphere to encase the player's head

## Changelog

Authored by: Clauda & Sandrine

```md
Assets:
- [Steel Sphere] A wide steel sphere to encase the wearer's head, blinding them and distorting hearing and speech
- [Steel Spheres] Steel mittens to be worn over the character's hands, rendering them useless
```

## Checklist


ClaudiaMia (commented):
Things I noticed with the steel sphere mittens:
In some poses the look a bit detached from the wrist ring to me in a weird way because sometimes the hand is still leaking out, e.g.
image

But there are other poses/instances. Not necessarily a blocker, but something I noticed. Also it is hard to align the left/right lighting in different poses, but  I do not think we can do much about that one.

Edit: Now I noticed, I feel the spheres are covering the middle of the hands, but somehow misaligned with the arm axis.

ClaudiaMia (changes requested):
The mittens are missing a preview image.

Also something I noticed already with the towel, but forgot to mention.
I do not think it is a good idea to name assets exactly the same, e.g. towel wrap or steel sphere. 

The preview image should make it obvious enough, but from the name you will not be able to quickly see what is what in the wardrobe when you wear both.

So as a suggestion: Why not call them something like: steel sphere mittens and steel sphere hood ? (or omitting "sphere" if you feel it gets too long)

ClaudiaMia (approved):

PR #305, opened 2024-11-27
[ADD] Serving tray

SandrinePDR:
### References

_None_

## About The Pull Request

Added a tray that can either be empty or hold a tea service. Further contents can be added, as well as alternate ways of holding it.

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Tray] New serving tray asset
```

## Checklist


ClaudiaMia (commented):
Great one!

How about making the tray colorable, too? I pushed a commit, please have a look.

ClaudiaMia (commented):
Reverted the last commit as it violated our process.

General review:
Tested the asset and it works and looks well.

Two findings from my side:

1. A user would expect that the tray can be colored, too. I see no reason why it should not be colorable. But it is in the end an artistic decision if it should be colorable by the asset contributor, so exceptions could be possible. 
Generally, our vision for Pandora is that assets can be colored as good as possible, so they should ideally not come pre-colored with no customization possible.

2. Personally, I wondered how the tray is held to the body despite gravity. It looked to me like it would flip over very easily, as there is just one chain on the front edges (if you look from the back side, it was quite noticeable to me). A possible improvement could be to have another chain on the backside around the hips.

ClaudiaMia:
These were just my personal thoughts. I also asked Clare to take over the actual reviewing and merge decision for this asset, as it feels weird for me to be the reviewer in this case, as I believe my actions lead to me no longer being a neutral and trusted party for the sake of a review. Again, I am sorry for my actions.

src/assets/accessories/tray/tray.asset.ts
```diff
@@ -44,6 +44,10 @@ DefineAsset({
 			],
 		},
 	},
+	chat: {
+		actionAdd: 'SOURCE_CHARACTER placed a serviing tray over TARGET_CHARACTER_DYNAMIC_POSSESSIVE neck.',
```
ClaudiaMia, line 48: typo in "serving"

ClaudiaMia (changes requested):

src/assets/accessories/tray/tray.asset.ts
SandrinePDR: Grmpf

ClaudiaMia: Soooorrryyy~~~

ClaudiaMia (approved):

Jomshir98 (approved):

PR #310, opened 2024-12-05
[FIX] Add missing point types to point templates

Jomshir98:
## References

_None_

## About The Pull Request

Some points templates were missing point types on some newer points.
This PR adds those types, as well as adding a check against this problem going forward.

## Changelog

Authored by: Clare & Claudia

```md
Fixes:
- Fixed some mesh vertices were missing correct type assignments. This will marginally reduce memory usage and might fix some very rare graphical glitches (so rare we didn't find exact example).
```

## Checklist


ClaudiaMia (approved):
Approved!

PR #320, opened 2024-12-17
[ADD] Body/legs split for the body template

Jomshir98:
## References

ref #213 (partially replaces)

## About The Pull Request

This PR adds a body/legs point type split for the "body" template.
This PR does not split any actual layers, merely adds the new point type so it can be used in the future. Transition was solved by automatically adding the new point types to existing layers that included `body` layer.

## Changelog

```md
Assets:
- Split the `body` template points into body and legs. This is not the actual planned body/legs split - only preparation for it.
Note, that as a result you will now, during asset creation, need to use `body,bodyarm,bodylegs,legs` as a point filter where you would use `body,bodyarm` previously.
```

## Checklist


ClaudiaMia (approved):
LGTM!

PR #321, opened 2024-12-18
[FIX] reworked cutting approach of the latex catsuit to remove artifacts

ClaudiaMia:
## References

_None_

## About The Pull Request

This PR changes the approach on how the latex catsuit's images are cut into different images and stitched back together, using point based cuts, rather than layering two bordering images together, as that created seam-like line artifacts on the arms and legs. This change approach fixes those.

Please not that I only changed the cutting approach for the suit layers, not for the reflection layers. The reflection layer still use the cutting approach that creates seams/line, but such a seam is only visible on the left butt cheek in the back view and it is barely noticeable, so I did not change anything for the reflection layers.

## Changelog

Authored by: Claudia

```md
Assets:
- [Latex Catsuit] Removed the seams/lines that were noticeable on the arms and legs when zooming in
```

## Checklist


SandrinePDR (approved):
LGTM!
Thank you so much for the rework @ClaudiaMia & @Jomshir98

Jomshir98 (approved):

PR #322, opened 2024-12-18
[ADD] new attribute for restricted covered hands and large asset attribute update

ClaudiaMia:
## References

_None_

## About The Pull Request

- Adds a new attribute `Hand_restricting_cover` that filled a gap we had, to complement the existing attribute `Hand_cover` that was too generic for all use cases.
- Rephrased some existing attributes and reordered them.
- Adjusted all handhelds as well as restraints such as mittens, armbinder, and straitjacket to use the available attributes better.

## Changelog

Authored by: Clare & Claudia

```md
Assets:
- Adds a new attribute `Hand_restricting_cover` complementing the existing attribute `Hand_cover` that was too generic for all use cases
- Adjusted several assets to use the available attributes more appropriately and realistically: All Handhelds, the finger nails, the armbinder, both mittens, and the straitjacket
```

## Checklist


SandrinePDR (approved):
LGTM!
Just two remarks, perhaps for later:
* Do we still require 'Mittens'? It seems to be replaced almost everywhere
* It would be nice, if we get some inheritance here. So 'Hand_restricting_cover' would include 'Hand_cover' aso. That would make the files less bloated from mops.

ClaudiaMia:
> LGTM! Just two remarks, perhaps for later:
> 
> * Do we still require 'Mittens'? It seems to be replaced almost everywhere
> * It would be nice, if we get some inheritance here. So 'Hand_restricting_cover' would include 'Hand_cover' aso. That would make the files less bloated from mops.

Indeed, I was actually almost done with removing the Mittens attribute everywhere, but then I stumbled over the steel spheres. Layered bondage of first mittens and then steel spheres on top might be nice, so I still needed the Mittens attribute to hide them under the spheres, as any other attribute would hide something that cover more than the hands, possibly. In the end I then decided to just keep it.

Yes, my idea was that 'Hand_restricting_cover' would not be used in parallel to 'Hand_cover', but as an alternative variant of it, still being used for something that also covers the hands. But yeah, there is no automatism or any checks, but I am not sure if it would be viable to do that or even needed.

Well, with the current amount of assets we can easily rework things again in the future, if we have a better idea later~

Jomshir98 (approved):
LGTM

PR #324, opened 2024-12-19
[ADD] Present

SandrinePDR:
## References

_None_

## About The Pull Request

Added a new accessory "Present" that can be extended to various kinds of gifts later. As it is the season at the moment, a Christmas present is the first option. Merry Christmas to all of you.

## Changelog

Authored by: Sandrine

```md
Assets:
- [Present] New asset
```

## Checklist


src/assets/room_devices/presents/presents.asset.ts
```diff
@@ -0,0 +1,48 @@
+DefineRoomDeviceAsset({
+	name: 'Christmas Presents',
```
ClaudiaMia, line 2: I am a bit on the fence having a not-colorable christmas-only present room item, but the effort would be high, and christmas is soon, so let's just go with it.

We can still change it at any time in the future.

ClaudiaMia (changes requested):
Looks veeery nice!
image

ClaudiaMia (approved):

PR #335, opened 2025-01-07
[ADD] Jute rope assets

Shikifet:
## References

_None_

## About The Pull Request

Add a new set of assets representing some shibari ties with multiple combinations.
Assets work between them, for example you can tie wrists from arm asset to crotch rope asset. To make them interact properly, order should match body position (arm over crotch, and crotch over leg), although most options work independently of the order.

## Changelog

Authored by: Shikifet

```md
Assets:
- [Arm Jute Ropes] New asset with close to 100 possible configurations
- [Crotch Jute Ropes] New asset with 18 possible configurations
- [Leg Jute Ropes] New asset with 40 possible configurations
```

## Checklist


src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
```diff
@@ -0,0 +1,155 @@
...
+	name: 'Crotch jute ropes',
+	size: 'small',
+	graphics: 'graphics.json',
+	colorization: {
+		rope: {
+			name: 'Rope',
+			default: '#D7AC4D',
+		},
+	},
+	attributes: {
```
ClaudiaMia, line 11: This asset currently provides no attributes and can therefore not be found via any of the icon based filter categories in the wardrobe. Myself, I do not typically filter for a word, but use the buttons.

Now, it is tricky to find the right home for this one, but I guess it is more legs then arms, so maybe same as the leg ropes:

```
		provides: [
			'Restraint',
			'Restraint_legs',
		],
```

But you can also pick something else, e.g. the chastity category, if you find it more suitable for finding this.

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
```diff
@@ -0,0 +1,155 @@
...
+					properties: {
+						stateFlags: {
+							provides: [
+								'thigs_free'
+							]
+						}
+					}
+				},
+				{
+					id: 'attached',
```
ClaudiaMia, line 122: When you select attached for the thighs module, two more ropes connect legs and crotch ropes on the front side. The back side does not seem to be connected. Is that intentional or is there some layer missing?

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
```diff
@@ -0,0 +1,311 @@
...
+					default: true,
+				},
+				{
+					id: 'linked',
+					name: 'Linked',
+				},
+			]
+		},
+	},
+	effects: {
```
ClaudiaMia, line 286: This is likely a copy and paste mistake, as there seems to be no reason why this blocks hands.

src/assets/arm_restraints/arm_jute_ropes/arm_jute_ropes.asset.ts
```diff
@@ -0,0 +1,270 @@
+DefineAsset({
+	name: 'Arm jute ropes',
```
ClaudiaMia, line 2: Easy to miss, but so far for the asset names specifically, each word of all other assets start with a capital character.

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
```diff
@@ -0,0 +1,311 @@
+DefineAsset({
+	name: 'Leg jute ropes',
```
ClaudiaMia, line 2: Easy to miss, but so far for the asset names specifically, each word of all other assets start with a capital character.

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
```diff
@@ -0,0 +1,155 @@
+DefineAsset({
+	name: 'Crotch jute ropes',
```
ClaudiaMia, line 2: Easy to miss, but so far for the asset names specifically, each word of all other assets start with a capital character.

ClaudiaMia (changes requested):
Really amazing work! Lots of options with these three assets!

One thing that took me a while to figure out was when I can tie the wrists module of the arm ropes to the crotch ropes. It too me a while to figure out that I can do it with "box low" "behind back" and "armbinder", but not the other two, which did not seem immediately obvious to me why this is the case.

Now I also understand why it took a while to figure out. When they are connected and I have the arms "behind neck", it seems confusing to me the crotch ropes won't allow me to add the waist rope (same with other arm ties)

Or why it forbids me to have an open crotch area, when the arms are in an armbinder tie. Maybe I am overlooking something about stability and how ropes would pull and how it would look.

Just something that made it quite a bit tricky to attach the two assets to each other as it required quite specific settings over several modules on both assets.

Otherwise just two minor things I found.

Really awesome and fun-looking work!~

Shikifet:
Thanks! I wanted to give people lots of choices, so everyone can find what better suits their playstyle

> One thing that took me a while to figure out was when I can tie the wrists module of the arm ropes to the crotch ropes. It took me a while to figure out that I can do it with "box low" "behind back" and "armbinder", but not the other two, which did not seem immediately obvious to me why this is the case.

I may add in the future layers for box normal and box high attached to crotch. For now it felt a bit weird so I chose to block them instead.

> Or why it forbids me to have an open crotch area, when the arms are in an armbinder tie. Maybe I am overlooking something about stability and how ropes would pull and how it would look.

You should be able to have open crotch with armbinder. Probably you had wrists of armbinder attached to vulva rope, and needed to be removed before changing to open crotch.

> Just something that made it quite a bit tricky to attach the two assets to each other as it required quite specific settings over several modules on both assets. Not saying you have to change anything here, just giving feedback on how my experience was.

This was something made on purpose. I wanted the more complex figures to be a bit tricky to achieve, so people doesn't just go to the full complex tie from the begining, and instead they "learn" the ropes on its way. on different plays, or even discovering them on other peoples

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
Shikifet: Added, with also chastity category as suggested

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
Shikifet: Didn't have plans for back to also be attached, but as it may lead to confussion, I added it

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
Shikifet: You are right, hands shouldn't be blocked here

ClaudiaMia (approved):

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
```diff
@@ -0,0 +1,305 @@
...
+					default: true,
+				},
+				{
+					id: 'linked',
+					name: 'Linked',
+				},
+			],
+		},
+	},
+	blockSelfAddRemove: true,
```
Jomshir98, line 286: Does it make sense not to be able to self-tie your legs? Or is this a mistake by copying config of arms to legs?

src/assets/arm_restraints/arm_jute_ropes/arm_jute_ropes.asset.ts
```diff
@@ -0,0 +1,270 @@
...
+						},
+					},
+				},
+			],
+		},
+	},
+	effects: {
+		blockHands: true,
+	},
+	blockSelfAddRemove: true,
```
Jomshir98, line 251: With this combination it means that (if "Bound usage" is allowed on the item):
- You cannot put the rope on yourself
- You cannot remove the rope from yourself
- You still _can_ change any module variant to a different one

I find this combination really weird, as some ties really feel like they are equivalent to untying the rope and re-tying it in a completely different way.

I see these ways this could be possibly improved (not all at once, just possible approaches):
- Add `interactionType: ItemInteractionType.ADD_REMOVE,` to modules that should be treated equivalently to the item being first removed, then modified and then added again (it basically adds same requirements for changing the module's state as there are for removing or adding the item in the first place).
- Add the `blockSelfModules: ['...', '...']` effect, listing modules that shouldn't be self-modifiable.
- Remove the `blockSelfAddRemove` property, meaning you could remove and add the item on yourself (unless your hands are already bound and the item is set not to allow bound usage). I feel this one could make sense even with some ties being rather strict - while it might be hard to get into them, with careful preparation you likely could get into these as self-bondage, even if definitely not as high quality as a rigger tying them.
- Move `blockSelfAddRemove` to only some module variants that are stricter (as opposed to removing it) - item will not be self-add/remove-able if at least one selected variant has this property.

Note: We will likely be replacing `blockSelfAddRemove` with only `blockSelfAdd` soon, to give the choice of the item being removable to the "Bound usage" setting.

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
```diff
@@ -0,0 +1,305 @@
...
+									legs: 'kneeling',
+									bones: {
+										leg_r: 5,
+										leg_l: 5,
+									},
+								},
+							],
+						},
+					},
+					default: true,
```
Jomshir98, line 173: Nitpicky suggestion (fully your choice, just giving my thoughts): Should the default be two independently tied rope loops? I feel it makes a bit more sense for default configuration to only be one of them - the one you are more likely to tie first (so likely the ankle one). And it is only a default anyway, one which you can change even before actually spawning the item.

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
```diff
@@ -0,0 +1,158 @@
...
+			name: 'Hips',
+			variants: [
+				{
+					id: 'none',
+					name: 'None',
+					default: true,
+					properties: {
+						stateFlags: {
+							requires: {
+								thigs_free: 'Rope over hips not tied to thighs',
```
Jomshir98, line 99: Nitpick: Depending on how other requirements are updated, it might be nice to reword for consistency.

src/attributes.ts
```diff
@@ -644,6 +644,19 @@ const ATTRIBUTES_DEFINITION_BASE = {
...
 	},
+	// Jute Rope
+	Waist_rope_anchor_point: {
+		name: 'Rope around waist',
+		description: 'A rope around the waist',
+	},
+	Vulva_rope_anchor_point: {
+		name: 'Crotch rope link',
+		description: 'A rope linking crotch rope to tied hands',
+	},
```
Jomshir98, line 655: Suggested wording that I feel better matches the current usage:
Name: Crotch rope
Description: A crotch rope that a linking rope can be tied to


Context in which it is used currently (when trying to select the option requiring this, not yet having it selected):
> The arm jute ropes requires "A rope linking crotch rope to tied hands" (must be worn under the arm jute ropes).

After suggestion:
> The arm jute ropes requires "A crotch rope that a linking rope can be tied to" (must be worn under the arm jute ropes).

src/attributes.ts
```diff
@@ -644,6 +644,19 @@ const ATTRIBUTES_DEFINITION_BASE = {
...
+	Thight_rope_anchor_point: {
+		name: 'Crotch rope link',
+		description: 'A rope linking crotch rope to tied thighs',
+	},
```
Jomshir98, line 659: Suggested wording that I feel better matches the current usage:
Name:  Thighs rope
Description: A rope over thighs that a linking rope can be tied to

Context in which it is used currently (when trying to select the option requiring this, not yet having it selected):
> The crotch jute ropes requires "A rope linking crotch rope to tied thighs" (must be worn under the crotch jute ropes).

After suggestion:
> The crotch jute ropes requires "A rope over thighs that a linking rope can be tied to" (must be worn under the crotch jute ropes).

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
```diff
@@ -0,0 +1,158 @@
...
+							provides: [
+								'Vulva_cover',
+								'Vulva_rope_anchor_point',
+							],
+							covers: [
+								'Vulva_item',
+							],
+						},
+					},
+				},
```
Jomshir98, line 85: Thinking about this covering any item inside vulva...
On one hand it can make sense if the rope is tied tightly... on the other hand it slightly limits RP possibilities, like slipping something (even a small egg, once we have that) inside while the rope is tied.
Perhaps it would be more RP friendly if the `covers` was removed, as then it is much easier to RP sliding the rope aside or temporarily loosening it a bit?
It also doesn't really reduce strictness of the item, as to remove any item underneath you would need free hands in the first place (at which point you can remove this item yourself), so removing this enforcement would only increase what the rigger can choose.
The `Vulva_cover` provide should definitely be kept, as that is important for forcing items ordered under this one not to clip.

Jomshir98 (changes requested):
Hi! The asset looks really great and I'm looking forward to getting it in, thank you!

Claudia already reviewed it from graphics side of things, so I am focusing on reviewing the technical and user-friendliness parts.
Overall I really like the assets and I feel that while I have several notes noted below, they are just details for polishing it up even more.

Feel free to ask if you have any questions or disagree with anything I wrote!

src/assets/arm_restraints/arm_jute_ropes/arm_jute_ropes.asset.ts
```diff
@@ -0,0 +1,270 @@
...
+						stateFlags: {
+							requires: {
+								uncinched: 'Remove cinches in Cinch module',
+								//untied_shoulder: 'Remove Extra rope in Chest module'
+							},
+						},
+						attributes: {
+							requires: [
+								'!Waist_rope_anchor_point',
+							],
```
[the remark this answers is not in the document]
Shikifet: The 'Waist Rope' of 'Crotch Jute Rope' asset was created mainly to give aestethic for 'Box Tie' and 'Armbinder' poses, but "Wrists Behind Neck", draws its own rope right over "Waist Rope", which was not pleasant as both mixed.
As I'm unsure if there is a way to tell the 'Crotch Jute Rope' asset (placed below the 'Arms Jute Rope') to not draw the layer, because it should be overriden by the upper one, I chose to block it for now.
Maybe in the future I'll change the layers of crotch 'Waist Rope' to match somehow the one of 'Wrists Behind Neck' one

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
Shikifet: Not nitpicky at all, and RP wise it has more sense

src/assets/panties/crotch_jute_rope/crotch_jute_rope.asset.ts
Shikifet: You are right, added it thinking in being restricting, but as you point out, only restricts the rigger

src/assets/arm_restraints/arm_jute_ropes/arm_jute_ropes.asset.ts
Shikifet: I get it, I saw them in another asset and thought it worked slightly different as you explained now. I opted for removing the property, as indeed I want to be able to use them on selfbondage.

src/assets/foot_restraints/leg_jute_ropes/leg_jute_ropes.asset.ts
Shikifet: A mistake by copying from arms config

src/assets/arm_restraints/arm_jute_ropes/arm_jute_ropes.asset.ts
```diff
@@ -0,0 +1,270 @@
...
+							},
+							armsOrder: {
+								upper: 'right',
+							},
+						},
+						stateFlags: {
+							requires: {
+								//uncinched: 'Remove cinches in Cinch module'
+							},
+						},
```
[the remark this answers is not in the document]
Shikifet: yes, the same one than previous comment

Jomshir98 (approved):

PR #338, opened 2025-01-12
[ADD] paper note asset and eyes 3 & 4 shine changes

ClaudiaMia:
## References

_None_

## About The Pull Request

Made the shine of eyes3 and eyes4 assets move as well and added a new asset.

## Changelog

Authored by: Claudia

```md
Assets:
- [Paper Note] Added a new asset that can be held in each hand or be kept somewhere else on the character
- [Eyes 3, Eyes 4] Changed the shine layer to now also move with the according eye states
- [Toolbox, Pocket dimension] Simplified names (from "Room Construction Tools" and "Pocket dimension [PoC]")
- Fixed issues with the custom chat messages of most handhelds
```

## Checklist


ClaudiaMia:
I do not get these four warnings from the validation `Layer filters for point types, but all points match anyway. Remove the unnecessary filter.`

I absolutely need to filter for point types there so it works like I want it.

Jomshir98:
> I do not get these four warnings from the validation `Layer filters for point types, but all points match anyway. Remove the unnecessary filter.`
> 
> I absolutely need to filter for point types there so it works like I want it.

Yes, you are correct. The check wasn't correctly accounting for mirrored points (points in point template that get generated - aren't explicitly part of the template). I updated the logic and pushed a fix on this branch.

ClaudiaMia:
Thank you!~ <3

src/assets/handhelds/paper_note/paper_note.asset.ts
```diff
@@ -0,0 +1,101 @@
...
+					},
+				},
+			],
+		},
+	},
+	chat: {
+		actionAdd: 'SOURCE_CHARACTER put a paper note into TARGET_CHARACTER_DYNAMIC_POSSESSIVE hand.',
+		actionRemove: 'SOURCE_CHARACTER removed the paper note from TARGET_CHARACTER_DYNAMIC_POSSESSIVE hand.',
+		actionAddCreate: 'SOURCE_CHARACTER took out a paper note.',
+		actionRemoveDelete: 'SOURCE_CHARACTER disposed of the paper note.',
```
Jomshir98, line 85: The create/delete actions don't quite feel correct. Trying using a character to spawn a new note on a different character and see what happens.

src/assets/handhelds/paper_note/paper_note.asset.ts
```diff
@@ -0,0 +1,101 @@
...
+										rotation: 'up',
+									},
+								},
+								{
+									rightArm: {
+										rotation: 'down',
+									},
+								},
+							],
+						},
```
Jomshir98, line 52: This can be simplified a lot (similarly the one below):
```suggestion
						poseLimits: {
							rightArm: {
								fingers: 'fist',
								rotation: ['up', 'down'],
							},
						},
```

ClaudiaMia: All instances on all handhelds improved accordingly.

src/assets/handhelds/paper_note/paper_note.asset.ts
ClaudiaMia: Fixed

Jomshir98 (approved):
LGTM!

PR #356, opened 2025-02-03
[ADD] Tails and tail plugs

SandrinePDR:
## References

_None_

## About The Pull Request

Added several tail body parts for non human players and three variants of an anal plug that are decorated with artificial fur

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Body] Added several variants of tails as part of the body
- [Tail plug] Added new anal plug with three tail extensions
```

## Checklist


ClaudiaMia (changes requested):
Impressive work!~

Had just one finding that felt unintended to me.

ClaudiaMia (commented):
Just changing the priority is not enough. You also need to change the order of the layers, or otherwise the stripes will not correctly show from the front.

That is why I pasted the full tested code in my initial comment.

ClaudiaMia (approved):

src/assets/body/tail/tail.asset.ts
```diff
@@ -0,0 +1,129 @@
...
+	preview: null,
+	attributes: {
+		provides: [
+			'Tail',
+		],
+	},
+	modules: {
+		tailType: {
+			type: 'typed',
+			name: 'Tail Type',
```
Jomshir98, line 29: [suggestion, optional] Here again I wonder whether all these tails should be a single asset. Does it make sense to switch between the options while wearing it? Do they have a lot in common for this to make sense? I would say the thin, normal and thick options do... but what about the rest?
See my note on the plug variant for more details.

src/assets/body/tail/tail.asset.ts
```diff
@@ -0,0 +1,129 @@
...
+				},
+				{
+					id: 'thick',
+					name: 'Thick',
+				},
+			],
+		},
+		withStripes: {
+			type: 'typed',
+			name: 'With Stripes',
```
Jomshir98, line 56: Not all these options work with all tail types. This could be solved by splitting the asset into multiple assets (mainly looking at the bunny tail, which I think would make sense as a separate asset with far fewer modules), or by using state flags.
This applies to all the modules (other than "Tail Type")

Jomshir98 (changes requested):
I have several notes to this. Overall nice addition, but I am not quite sure these should really be only two assets. See the individual notes for more in-depth thoughts (don't forget to click "view more" to see all of them).

src/icons/tail.svg
```diff
@@ -0,0 +1,25 @@
...
...px2mmqYUrSOQFzXhXKpxDjn2pG6V1JKJKUELoR/7rnn&#10;qhhE0ToC+eXLL8tsNguUUh2mWGoSiQQAJCHEp5R6AKAEomgdgXz6wx/GCKEEQuiIm52klDwMw9Eg&#10;CFy1BBQtJZCuri5sxWIZSql5FAKRIecjJbWNrmg1gbywbh0qFYu6EIJNtVgx2kUXQoRhELimaYZq&#10;CShaSiALFy2imJAkABzRwIWJWizf97yh4vi4EoiitQSSz+epYRi9CKHYkbx+YpicwISM3vfAAyoG&#10;UbSWQMIwNGOxmK1pGhypQDjn9TAMndOXLlUZLEVrCcTQ9QwCaEdH0S3FOa87jlN2ajUlEEVrCQQQ&#10;SiGM++AI++k559L3/TKjdNf1f/mXKgZRtI5Avvkv/6JJKXs0xmyCj+xXF0L4juO8Wi6Xh9XtV7SU&#10;QAzThIRtZwkhSThCDysIAvA8b8+Lzz9fVLdf0VIC2bFjh4jF44xpGjvSEMTzvFK1Wt0za+5cVeau&#10;aC2BXH755cSyLI8Q4hxh/AG+7w/yMHxp06ZN6nxCRWsJZMmSJX61UtnpuW6l+Yi1KViP6sjIyMZC&#10;obD/f95wg5qHpWgtgay59VZY98ILu0ZGR9cJIfwpxh68OD6+r1IuP3jNRRcNqluvaDmBXPupT0nG&#10;2M7A918YKxSKnPPDsgJhGMpqtToyNjb2Hy+uW/fMX3/967669YrDYdoNr77nnnvE2Wedtce27TZM&#10;yELDMOibbRr6vi+r1epIoVD4/muvvXaXaVnbr/nkJ5V7pWhNgQAAnHPOOS5lbFMikbAAoEPTNBMA&#10;SKQTeQDkum61UCgMjoyM3LJly5a7k8nkpov/9E9V9krR2gJ54IEH+MIFCyqVSmWDaVkFz/MKvu+n&#10;giBwHcfhnucN1Wq1jfv27XusVCp9++mnn74vFovtvuDCC5XlUEyJ6T6wgPzgttv0XD6f6e3pOcl1&#10;3awESFJKS5VyefP4+Hjxsg9+cGTNmjX82muvVXVXihNOIJO4+atfpe35vLQsCy6/4grlSikUCoVC&#10;oVAoFAqFQqFQKBQKhUKhUCgUCoVCoVAoFAqFQqFQKBQKhUKhUCgUCoVCEfH/AViQVC6hlIguAAAA&#10;AElFTkSuQmCC&#10;"
```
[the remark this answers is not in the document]
SandrinePDR: Well that was what came out of a png to svg tool. But you are right, listing this under the fantasy icon is better

SandrinePDR:
Ready for review

ClaudiaMia (changes requested):
I tested the new changes and found some more things:
1. We wanted to add the fantasy items filter button to the body tab I believe, so your nice work can easily be found. You can do that by adding also `body` to the "Fantasy" attribute's "tabs" array in attributes.ts#250
2. The "Cat Tail" body item does not show a preview in the body tab.
3. The "Cat Tail" has a "Module: With Stripes" that does nothing and feels to me like a part that was forgotten to be removed, as the next module does the same.

Jomshir98 (approved):
From code view looks mostly good to me now. Seems you missed one note about not including `required: false` in the bodypart definition, but good otherwise.

ClaudiaMia (approved):
Yay! What a nyanderful work!~

PR #359, opened 2025-02-04
[ADD] Glove add-on for the latex catsuit

SandrinePDR:
## References

_None_

## About The Pull Request

Added latex gloves as an add-on to the already existing catsuit

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Catsuit] Added gloves as an add-on
```

## Checklist


ClaudiaMia (commented):
I found two things when I tested it.

For once I wondered why I cannot have gloves in the short variant, too? The graphics look to me like they are long enough to nicely work also in the combination of short length plus gloves (other states make not too much sense for sure, but with state flags this one combination could be supported), leaving the legs uncovered.

The other thing I noticed is actually the reverse issue: When the length is long, just the "glove" state makes the long legs suddenly disappear , which seems not logical and more like a bug to me.

SandrinePDR:
I changed the two points you remarked

ClaudiaMia (approved):
Very very sweet!

Thank you!~

src/assets/dresses/latex_catsuit/latex_catsuit.asset.ts
```diff
@@ -82,7 +82,7 @@ DefineAsset({
 		},
 		addOns: {
 			type: 'typed',
-			name: 'Add-Ons',
+			name: 'Add-Ons (socks only with long variant)',
```
Jomshir98, line 85: Does this mean, that the two "socks" options only make sense with the long variant and never with short variant?
If that is the case, this should be done the same way other assets block invalid combinations - using state flags. That way things are standardized instead of random notes in the name.

For example how to do this, have a look at "Arm Jute Ropes".

Jomshir98 (changes requested):
Looks mostly good. I just have a problem with the random note in the module name instead of using state flags (see my code comment).

src/assets/dresses/latex_catsuit/latex_catsuit.asset.ts
SandrinePDR: You never stop learning...

Jomshir98 (changes requested):
Testing your change wouldn't have been a bad idea. You might have noticed that this is broken then.
I added a new validation that should point you in the right direction.
This new validation also reports a weirdness on the "Connecting Chains" assets that you should definitely look into (as that is your asset, you would know how it was originally meant).

SandrinePDR:
Tbh, I cannot remember adding the StateFlag attribute to the connecting Chain asset, so I removed it entirely.

Jomshir98 (approved):

PR #360, opened 2025-02-05
[ADD] Rope and Bamboo gags

Shikifet:
## References

_None_

## About The Pull Request

Add 2 new gags assets: a simple bamboo gag and a rope gag with 3 options (just the rope over mouth, with a knot like a ballgag or coiled making a large bit gag instead)

## Changelog

Authored by: Shikifet

```md
Assets:
- [Bamboo Gag] Added a new "Bamboo Gag" asset
- [Jute Rope Gag] Added a new "Jute Rope Gag" asset
```

## Checklist


src/assets/gags/jute_rope_gag/jute_rope_gag.asset.ts
```diff
@@ -0,0 +1,113 @@
...
+							throatBreath: 3,
+							coherency: 5,
+							stimulus: 2,
+						},
+					},
+				},
+			],
+		},
+	},
+	effects: {
```
ClaudiaMia, line 86: This effect definition here is redundant as you define an individual effect per module state.

src/assets/gags/jute_rope_gag/jute_rope_gag.asset.ts
```diff
@@ -0,0 +1,113 @@
...
+			'Restraint',
+			'Restraint_mouth',
+			'Mouth_item',
+			'Mouth_insert',
+			'Mouth_cover',
+		],
+		requires: [
+			'Mouth_open_wide',
+			'!Mouth_protruding',
+			'!Mouth_tongue_out',
```
ClaudiaMia, line 25: More of an open question, as I cannot say: Would you say it is also impossible to sneak out the tongue in the rope state of the gag, or only in the knot and bit config?

If not, you can also remove it here and add this specific requirement to the two module states specifically.

ClaudiaMia (changes requested):
Wow! Very nice!~

Just two minor comments from my side.

src/assets/gags/jute_rope_gag/jute_rope_gag.asset.ts
Shikifet: I don't believe it's impossible, so moved it to both module states instead

ClaudiaMia (approved):
<3

Jomshir98 (approved):
LGTM!

PR #367, opened 2025-02-12
[CHANGE] Chain leash

SandrinePDR:
## References

_None_

## About The Pull Request

Split the existing Chain Leash asset (after discussing it with @ClaudiaMia ) into separate layers for chain and handle and added two new positions, so that people who are kneeling can be leashed around by standing players 

## Changelog

Authored by: Claudia (original asset), Taja & Sandrine

```md
Assets:
- [Chain Leash] Separate layers for chain and handle, two new leash options
```

## Checklist


ClaudiaMia (commented):
I feel the quality of the new state's images are not ideal, as the chain is quite skewed and wavy at the end.

Also I feel the handle there might need to be a bit longer, as it feels almost only half as long as the handle when the chain is hanging straight down.

But not sure if it is worth doing something about. We can also merge it like this.

SandrinePDR:
> I feel the quality of the new state's images are not ideal, as the chain is quite skewed and wavy at the end.
> 
> Also I feel the handle there might need to be a bit longer, as it feels almost only half as long as the handle when the chain is hanging straight down.
> 
> But not sure if it is worth doing something about. We can also merge it like this.

We adjusted the handle length and also did some tweaks to the graphics, but I fear that's the best we can do at the moment.

ClaudiaMia (approved):

PR #373, opened 2025-02-14
[CHANGE] Interaction type for tail positions

SandrinePDR:
## References

_None_

## About The Pull Request

Changed the interaction type for the new body part tails so that their position can be changed even ion rooms that prohibit body changes 

## Changelog

Authored by: Sandrine

```md
Assets:
- [body/cat tail, body/ puppy tail] Changed the tail position to be an expression for quicker access and so that it can be changed even in rooms that block body modifications
```

## Checklist

### DIFF
--- a/src/assets/body/tail_cat/tail_cat.asset.ts
+++ b/src/assets/body/tail_cat/tail_cat.asset.ts
@@ -1,3 +1,5 @@
+import { ItemInteractionType } from 'pandora-common';
+
 DefineBodypart({
 	name: 'Cat Tail',
 	bodypart: 'tail',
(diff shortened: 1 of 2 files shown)


ClaudiaMia (commented):
This works, but there is an alternative:

Adding
`expression: 'Horizontal tail direction'`
and
`expression: 'Vertical tail direction'`

"If this is a typed module without explicitly set interactionType, then this being set to an expression makes interactionType default to ItemInteractionType.EXPRESSION_CHANGE."

So if you want that only the character with the tail can change their direction, you do not need any interaction type as an expression, but if you want others with permission to also be able to change it, then you keep the "ACCESS_ONLY" type as you added it.

As an expression, you can change it via the expression tab in the space and wardrobe, which acts as a quick menu (and the UI will later on be improved for sure to make it quicker).

Just telling you that this options exists. Not sure if you see a body part tail the same as eyes or lips.

ClaudiaMia (approved):

PR #375, opened 2025-02-16
[CHANGE] Tail plug layer

SandrinePDR:
## References

_None_

## About The Pull Request

Changed the graphical representation of the  tail plugs, so that they even can be used with tight fitting clothes. This is a small break in consistency but for rp-sake it's better this way.

## Changelog

Authored by: Sandrine

```md
Assets:
- [Bunny tail plug, Cat tail plug, Puppy tail plug] Changed the layer of the tail to always show
```

## Checklist


src/assets/toys/bunny_tail_plug/bunny_tail_plug.asset.ts
```diff
@@ -20,7 +20,7 @@ DefineAsset({
 			'Toy',
 		],
 		requires: [
-			'!Anus_cover',
```
ClaudiaMia, line 23: I understand why you want to remove this negative requirement from all three tail plugs, but it is problematic to do so.
Because now you can add a tail plug over a locked chastity belt.
The plug would technically be outside of the cover, and it is also not locked inside the belt, but it seems like it is inside, leading to an unclear and confusing situation.

If you want to enable roleplaying that every belt, bondage bag and tight clothing could always have a small gap where a tail could be lead to the outside from the plug inside (under it), the better solution is to rework these three assets to not be plugs one their own, but rather that they are tail addons (similar to the leash that requires a ring) that require another item (a `Anus_insert` item) to be under them in the wear order.

ClaudiaMia (changes requested):
I found some problem.

src/assets/toys/bunny_tail_plug/bunny_tail_plug.asset.ts
ClaudiaMia: Clare had one more idea, that could even be the best approach:

To do the same "hack" as for the male sex: Adding a similar module to the three plug tails that moves the tail graphics from `ABOVE_BODY` layer to the `OVERLAY` layer on demand.

ClaudiaMia (changes requested):
That is not a module that allows you to switch layers. Did you look at the male sex like I mentioned in my comment?

Always background priority is wrong, as that is not the expected behavior of how wardrobe ordering works. The default and realistic behavior is that your tail is under a coat when you wear one or that you at least can layer it freely and not be forced that the tail always peeks through all layers of clothing.

The special case you wanted for some RP scenarios is being able to switch this behavior to the background layer.

That requires a module.

I can take the PR over if you want help.

SandrinePDR:
> That is not a module that allows you to switch layers. Did you look at the male sex like I mentioned in my comment?
> 
> Always background priority is wrong, as that is not the expected behavior of how wardrobe ordering works. The default and realistic behavior is that your tail is under a coat when you wear one or that you at least can layer it freely and not be forced that the tail always peeks through all layers of clothing.
> 
> The special case you wanted for some RP scenarios is being able to switch this behavior to the background layer.
> 
> That requires a module.
> 
> I can take the PR over if you want help.

That would be awesome, as I'm short on time this week...

PR #377, opened 2025-02-17
[ADD] New underwear

SandrinePDR:
## References

_None_

## About The Pull Request

Added nine new variants of panties and an optional 'wet spot' for naughty girls

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Panties] 9 new underwear assets
- [Style1] Renamed to 'Plain Panties' and also added the 'wet spot' module that the new ones have
```

## Checklist


ClaudiaMia (changes requested):
This is an absolutely amazing work!

I did find some problems though, related to the assets not following the form of the body.
Underwear or clothes worn under other clothes that do not follow the body shape are a huge problem while we have no well working alpha masks.

Two issues:
- All panties are not tight around the crotch, but are rather hanging slightly down, which leads to them being visible as a line behind some other items that can be worn above underwear (e.g. jeans shorts or leggings).
- Some panties do not follow the body form well enough and stick out behind pretty much all other clothing. Those two are:
- boy briefs
- control briefs

PR #380, opened 2025-02-18
[ADD] Chopsticks Set

Shikifet:
## References

_None_

## About The Pull Request
Adds a set of chopsticks that can be used to strap the tongue (as gag) or the nipples (as toy).

## Changelog

Authored by: Shikifet

```md
Assets:
- [Chopsticks Gag] Added new asset "Chopsticks Gag"
- [Chopsticks Nipple Clamps] Added new asset "Chopsticks Nipple Clamps"
```

## Checklist


src/assets/gags/chopsticks_gag/chopsticks_gag.asset.ts
```diff
@@ -0,0 +1,58 @@
+DefineAsset({
+	name: 'Chopsticks Gag',
+	size: 'small',
+	requireFreeHandsToUseDefault: true,
```
ClaudiaMia, line 4: I believe setting this to `false` makes more sense in this case, as this is not a gag that is held in place by a strap or harness, so you could realistically slide it off forcefully by moving your head against something.

src/assets/toys/chopsticks_nipple_clamps/chopsticks_nipple_clamps.asset.ts
```diff
@@ -0,0 +1,87 @@
+DefineAsset({
+	name: 'Chopsticks Nipple Clamps',
+	size: 'small',
+	requireFreeHandsToUseDefault: true,
```
ClaudiaMia, line 4: I believe setting this to false makes more sense in this case, as you could realistically slide them off forcefully by moving your body against something.

ClaudiaMia (changes requested):
Wow! We only talked about yesterday and here you are with such amazing assets!

I just spotted two minor things, about the default of being able to struggle them off for RP purposes.

ClaudiaMia (approved):
Very great assets! Thank you!

PR #385, opened 2025-02-21
[ADD] Piercings

SandrinePDR:
## References

_None_

## About The Pull Request

Added a first set of piercings for nose, nipples, navel and clitoris. Each comes with several variants, including a cute ring for later use.

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Piercings/xxx] New assets for nose-, nipple-, navel- and clitoris-piercings
```

## Checklist


src/assets/piercings/clitoris/clitoris.asset.ts
```diff
@@ -0,0 +1,89 @@
...
+			name: 'Piercing',
+			default: '#BBBBBB',
+		},
+	},
+	// size:100, y:700, centered
+	preview: 'preview.png',
+	attributes: {
+		provides: [
+			'Piercing',
+			'Accessory',
```
ClaudiaMia, line 19: I am not sure if we should list these four piercings also under accessories, if we have a whole filter section just for them. As this slowly makes the accessory filter list too big.

src/assets/piercings/nipple/nipple.asset.ts
```diff
@@ -0,0 +1,119 @@
...
+				{
+					id: 'none',
+					name: 'None',
+					default: true,
+				},
+				{
+					id: 'double',
+					name: 'Double Stud',
+					properties: {
+						attributes: {
```
ClaudiaMia, line 31: In the other three assets, you provide the universal attributes on the top level, but here you provide them on the module level. I think it should be consistent. So it's better to move these out of the module level.

`Piercing_chainable` is a tricky one though and is for sure module specific. That said, since you allow the breast piercings to be one-sided, too, any future chain options attached to it would also need to support the option to be one-sided.

src/attributes.ts
```diff
@@ -665,6 +665,10 @@ const ATTRIBUTES_DEFINITION_BASE = {
 		name: 'A chainable armbinder',
 		description: 'An armbinder that is chainable to another item',
 	},
+	Piercing_chainable: {
```
ClaudiaMia, line 668: I am not sure this generic attribute will work for the purpose you intent, as it does not identify the location or type of the piercing uniquely.

I believe it would need several specific attributes, such as:
- clitoris_piercing_chainable
- navel_piercing_chainable
- nipple_piercing_chainable_left
- nipple__piercing_chainable_right
- nose_piercing_chainable

ClaudiaMia (changes requested):
Nice work! I do have some comments.

src/assets/piercings/clitoris/clitoris_ring.png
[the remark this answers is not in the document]
SandrinePDR: I discussed this already with taja and she likes it like this

src/attributes.ts
SandrinePDR: I commented it with //TODO as discussed

Jomshir98 (approved):
Code-wise looks fine. I will be fixing up the attributes before this is released.

PR #395, opened 2025-03-10
[ADD] More timer locks

ClaudiaMia:
## References

ref #915  

## About The Pull Request

-

## Changelog

Authored by: Claudia

```md
Assets:
- Changed the previous timer lock asset to now support 48 instead of 24 hours as maximum value
- Added two new timer locks that support a maximum value of 1 hour and 12 hours respectively
- Added two new timer locks that support a maximum value of 7 days and 1 year - both of these locks have the item limit "prevent" set per default, so no other character can use them or request usage out of the box. Users who want to make one of those locks available need to change the defaults in the "item limits" tab of the wardrobe.
- [Latex Corset] Fixed a clipping issue with larger bust sizes
```

## Checklist


Jomshir98 (approved):
LGTM!

PR #398, opened 2025-03-11
[ADD] Kigurumi mask

Shikifet:
## References

Adds a Kigurumi mask as a new asset. It can be used as a single mask or with a back piece, enclosing all head and hiding hair 'forcing' the use of wigs over it.

## About The Pull Request

It also updates hairstyle 12 to avoid a small glitch when using wigs over mask.

## Changelog

Authored by: Shikifet

```md
Assets:
- [Kigurumi Mask] Add new asset Kigurumi mask
- [Hairstyle 12] Update layers to avoid visual glitch while using Kigurumi mask
```

## Checklist


ClaudiaMia (changes requested):
Finished my review and have a few comments:

1. There seem to be no ears when you look from behind. Only when you look from the front.

2. Since the mask and helmet are bigger than the face, it leads to some visual glitches on some front hairs as it cuts them off somehow, but it is very minor so I feel that is okay.

What is more noticeable is that some back wigs are so small compared to the helmet that you can see the helmet bleed through at the top edge quite a bit. If you want to keep the mask this large, then I guess what could help is to have the ability to color the helmet in the color of the hair, which you can already do right now, but it is the same color for the face mask. So maybe mask and helmet should be able to be colored independently?

3. The black holes in the mask, are they supposed to show the color of the real hair underneath or are they black and uncolorable because they are supposed to be dark from the shadow in the holes? Just asking~

Otherwise very nice work that shows the power of the ease of reuse in Pandora, I guess~

Shikifet:
> Finished my review and have a few comments:
> 
> 1. There seem to be no ears when you look from behind. Only when you look from the front.
> 2. Since the mask and helmet are bigger than the face, it leads to some visual glitches on some front hairs as it cuts them off somehow, but it is very minor so I feel that is okay.
> 
> What is more noticeable is that some back wigs are so small compared to the helmet that you can see the helmet bleed through at the top edge quite a bit. If you want to keep the mask this large, then I guess what could help is to have the ability to color the helmet in the color of the hair, which you can already do right now, but it is the same color for the face mask. So maybe mask and helmet should be able to be colored independently?
> 
> 3. The black holes in the mask, are they supposed to show the color of the real hair underneath or are they black and uncolorable because they are supposed to be dark from the shadow in the holes? Just asking~
> 
> Otherwise very nice work that shows the power of the ease of reuse in Pandora, I guess~

Thanks for the review!

1. Yes, back ears were missing, so I added them.
 
2. Size of mask has become the most challenging in this asset, as I wanted it to be at least a bit bigger than face, so it could enclose the head (and having space to hide hair). In the end I left ears the same size so they could fit the best possible with current hairstyles while keeping the mask a bit bigger. 
For back side, I made it smaller, as the illusion is done on the lower part, so now it looks better. 

I know is not the optimal approach, but I think it keeps now a good balance. Maybe at some point in the future, it could be added a new type of module, like storage one, filtering by wigs and let them to be drawn on canvas resizing the wigs to match mask size for a perfect fit, similar to how room_devices works.

3. Also, I didn't make back holes colorable, because it was meant to represent the darkness of the holes. I imagine the hair is in some kind of hairnet inside the mask. In fact, in front mask its suppose to be some tape so wig can be fixed to it, but thought that making it colorable or giving another module was not necessary, as there are already lots of colors and modules.

And finally yes, there was a lot of reuse in this asset, and there could be more as I only chose one kind of eyes, mouth and nose, but felt that having too many options would be too much, and better focusing on other options, like different kind of lenses.

ClaudiaMia:
> > Finished my review and have a few comments:
> > 
> > 1. There seem to be no ears when you look from behind. Only when you look from the front.
> > 2. Since the mask and helmet are bigger than the face, it leads to some visual glitches on some front hairs as it cuts them off somehow, but it is very minor so I feel that is okay.
> > 
> > What is more noticeable is that some back wigs are so small compared to the helmet that you can see the helmet bleed through at the top edge quite a bit. If you want to keep the mask this large, then I guess what could help is to have the ability to color the helmet in the color of the hair, which you can already do right now, but it is the same color for the face mask. So maybe mask and helmet should be able to be colored independently?
> > 
> > 3. The black holes in the mask, are they supposed to show the color of the real hair underneath or are they black and uncolorable because they are supposed to be dark from the shadow in the holes? Just asking~
> > 
> > Otherwise very nice work that shows the power of the ease of reuse in Pandora, I guess~
> 
> Thanks for the review!
> 
> 1. Yes, back ears were missing, so I added them.
> 2. Size of mask has become the most challenging in this asset, as I wanted it to be at least a bit bigger than face, so it could enclose the head (and having space to hide hair). In the end I left ears the same size so they could fit the best possible with current hairstyles while keeping the mask a bit bigger.
>    For back side, I made it smaller, as the illusion is done on the lower part, so now it looks better.
> 
> I know is not the optimal approach, but I think it keeps now a good balance. Maybe at some point in the future, it could be added a new type of module, like storage one, filtering by wigs and let them to be drawn on canvas resizing the wigs to match mask size for a perfect fit, similar to how room_devices works.
> 
> 3. Also, I didn't make back holes colorable, because it was meant to represent the darkness of the holes. I imagine the hair is in some kind of hairnet inside the mask. In fact, in front mask its suppose to be some tape so wig can be fixed to it, but thought that making it colorable or giving another module was not necessary, as there are already lots of colors and modules.
> 
> And finally yes, there was a lot of reuse in this asset, and there could be more as I only chose one kind of eyes, mouth and nose, but felt that having too many options would be too much, and better focusing on other options, like different kind of lenses.

Great, but I still think it could make sense to allow coloring the helmet separately from the mask potentially? What do you think?

Shikifet:
> > > Finished my review and have a few comments:
> > > 
> > > 1. There seem to be no ears when you look from behind. Only when you look from the front.
> > > 2. Since the mask and helmet are bigger than the face, it leads to some visual glitches on some front hairs as it cuts them off somehow, but it is very minor so I feel that is okay.
> > > 
> > > What is more noticeable is that some back wigs are so small compared to the helmet that you can see the helmet bleed through at the top edge quite a bit. If you want to keep the mask this large, then I guess what could help is to have the ability to color the helmet in the color of the hair, which you can already do right now, but it is the same color for the face mask. So maybe mask and helmet should be able to be colored independently?
> > > 
> > > 3. The black holes in the mask, are they supposed to show the color of the real hair underneath or are they black and uncolorable because they are supposed to be dark from the shadow in the holes? Just asking~
> > > 
> > > Otherwise very nice work that shows the power of the ease of reuse in Pandora, I guess~
> > 
> > 
> > Thanks for the review!
> > 
> > 1. Yes, back ears were missing, so I added them.
> > 2. Size of mask has become the most challenging in this asset, as I wanted it to be at least a bit bigger than face, so it could enclose the head (and having space to hide hair). In the end I left ears the same size so they could fit the best possible with current hairstyles while keeping the mask a bit bigger.
> >    For back side, I made it smaller, as the illusion is done on the lower part, so now it looks better.
> > 
> > I know is not the optimal approach, but I think it keeps now a good balance. Maybe at some point in the future, it could be added a new type of module, like storage one, filtering by wigs and let them to be drawn on canvas resizing the wigs to match mask size for a perfect fit, similar to how room_devices works.
> > 
> > 3. Also, I didn't make back holes colorable, because it was meant to represent the darkness of the holes. I imagine the hair is in some kind of hairnet inside the mask. In fact, in front mask its suppose to be some tape so wig can be fixed to it, but thought that making it colorable or giving another module was not necessary, as there are already lots of colors and modules.
> > 
> > And finally yes, there was a lot of reuse in this asset, and there could be more as I only chose one kind of eyes, mouth and nose, but felt that having too many options would be too much, and better focusing on other options, like different kind of lenses.
> 
> Great, but I still think it could make sense to allow coloring the helmet separately from the mask potentially? What do you think?

It could avoid problems with future hairstyles, but kept the ears with the same colorization key of face plate although technically are also part of the helmet.

ClaudiaMia (approved):
Great work!~

PR #399, opened 2025-03-15
[ADD] Suspension Ring

Shikifet:
## References

_None_

## About The Pull Request

Add a new room device, representing a metal ring to do basic suspensions.

Some positions require Leg Jute Ropes with thighs tied whil others require legs not tied.
Warning: Please dont try to use outdoor.

## Changelog

Authored by: Shikifet

```md
Assets:
- [Suspension Ring] Added new room device "Suspension Ring"
- [Suspension Bamboo] Fix typos in notification
```

## Checklist


src/assets/room_devices/suspension_ring/chest_line_kneeling.png
ClaudiaMia, line 1: Is it intentional that this image not ending in a knot at the top end like the standing/center image does? If it is supposed to look like the end is hanging down on the back side, then it is not aligned ideally with the ring.
image

ClaudiaMia (changes requested):
Very very nice asset!

Just possibly one small finding and two thoughts.

And then I have some open thoughts:
I guess just tying it to the back  and backside of the thighs while the ring is fully frontal view might raise gravity/view related questions from users with less imagination, though I discussed it with Clare and we feel it is fully fine as one has a number of possibilities how to explain it to / see it oneself:
image

And then there is the matter of explaining the stretched out lower leg, something that is surely extremely physically straining. I do not think we will ever have a way/pose/view that lets us arc the leg correctly downward. They best we will be able to do is  frog tie one selected leg so that the raised leg ends at the knee visually. Another idea could be to add another rope from the ankle that is also tied to the ring to "explain" the pose.
But again, fully fine like this and no need to change anything.
image

src/assets/room_devices/suspension_ring/chest_line_kneeling.png
Shikifet: Not intentional, missing element both in kneeling and suspending pose. Updated them fixing it

ClaudiaMia (approved):
Looking good to me!~

Shikifet:
Regarding the stretched leg, as you have already figured out, the asset is meant to rise the leg, and then frogtie it. Idea is to frogtie only the suspended leg, but as that is not yet possible, it can work like that.

If some day both legs logic are splitted, then I'll add to the legs rope the possibility to frogtie only one of them, and to the suspension ring we can make it mandatory to have the leg frogtied.

PR #402, opened 2025-03-19
[FIX] Kigurumi mask

Shikifet:
## References

_None_

## About The Pull Request

Updates Kigurumi mask adding a module to hide or show fantasy ears while helmet is on and adding attribute to cover Mouth_items when mask is on

## Changelog

Authored by: Shikifet

```md
Assets:
- [Kigurumi mask] Add module to show/hide fantasy ears while helmet is on
- [Kigurumi mask] Add cover Mouth_item attribute while mask is on
```

## Checklist

### DIFF
--- a/src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
+++ b/src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
@@ -351,7 +351,7 @@ DefineAsset({
 						},
 						stateFlags: {
 							requires: {
-								helmet: 'Fantasy ears cannot be hidden without helmet',
+								helmet: 'Non-human ears cannot be hidden without helmet',
 							},
 						},
 					},
--- a/src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
+++ b/src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
@@ -333,7 +333,7 @@ DefineAsset({
 		},
 		fantasy_ears: {
 			type: 'typed',
-			name: 'Fantasy Ears',
+			name: 'Non-human Ears',
 			variants: [
 				{
 					id: 'visible',
--- a/src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
+++ b/src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
@@ -91,7 +91,6 @@ DefineAsset({
 								'Hair',
 								'Wig',
 								'Ears',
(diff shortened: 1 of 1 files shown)


src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
```diff
@@ -330,6 +331,33 @@ DefineAsset({
 				},
 			],
 		},
+		fantasy_ears: {
+			type: 'typed',
+			name: 'Fantasy Ears',
```
ClaudiaMia, line 336: While the attribute name is "fantasy_ears", the user does not know that. The not real pet ear assets are called "artificial" so far and the real pet ears are just called ears.

So "fantasy ears" is likely a term accurate enough to cover both, but I still worry that it might lead to questions/confusions on the user side.

My question is if we need this module at all? Why should the mask hide any ears that are peeking out behind the mask? If they are real, you likely want them to stay, and if they are artificial you could remove them if you do not want them.

ClaudiaMia: Hmm.. I guess it is akin to a helmet with holes for the real (pet) ears or not. So you have the option to look human.

Then I guess the term "Fantasy Ears" might be the best term we can come up with to not make people think that the module hides the ears of the mask.

Shikifet: You are right on name, maybe it should be call "Pet ears" instead for users?

Idea of module was to allow natural fantasy ears to poke out of the mask through some holes on top, because it felt uncomfortable for 'natural pets' having their ears forced inside the mask, as there is no accomodation for them like human ears have.

Ideal case would be that this module only works for natural and not artificial ears, but couldn't figure out if that was possible.

I asked to a couple pets in club if they would want their ears poking out of mask, and they both agreed, so thought this module would be the best solution.

src/assets/blindfolds/kigurumi_mask/kigurumi_mask.asset.ts
```diff
@@ -330,6 +331,33 @@ DefineAsset({
...
+					name: 'Hidden',
+					properties: {
+						attributes: {
+							hides: [
+								'Fantasy_ears',
+							],
+						},
+						stateFlags: {
+							requires: {
+								helmet: 'Fantasy ears cannot be hidden without helmet',
```
ClaudiaMia, line 354: Here, too~

ClaudiaMia (approved):
Thank you!~

PR #404, opened 2025-03-24
[ADD] Eye patch as a new accessory

SandrinePDR:
## References

_None_

## About The Pull Request

Added a new "eye patch" asset as an accessory

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [eye patch] New accessory
```

## Checklist


ClaudiaMia (changes requested):
Looks very nice!

I just want to note that it can make sense that the patch comes with a blind effect, as we also provide the option that closing one eye provides a partial effect, but it is optional.

I want to note that I think that there are surely people who believe that only an item covering both eyes should blind, as realistically, with one eye covered you just see less, but it's not darker. One can easily add a blind effect via mods as option.

But I am fine with leaving the blind effect, as a detail like that should be the asset makers' choice to a degree, as long as things do not got totally inconsistent.

But if the blind effect will stay enforced, the asset shall be changed to be a "blindfold" and not an "accessory", as that category is defined as primarily items that decrease the ability to see.

src/assets/accessories/eye_patch/eye_patch.asset.ts
```diff
@@ -34,6 +31,26 @@ DefineAsset({
 				},
 			],
 		},
+		blindness: {
+			type: 'typed',
+			name: 'Add Blind Eeffect',
```
ClaudiaMia, line 36: Typo spotted: "Eeffect" -> "Effect"

ClaudiaMia (approved):

PR #405, opened 2025-03-24
[ADD] Beauty spots

SandrinePDR:
## References

_None_

## About The Pull Request

Added beauty spots as a new face adornment.

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Beauty Marks] New accessory 
```

## Checklist


ClaudiaMia (changes requested):
I would have expected that this is another bodypart "skin" item? Is this supposed to be artificial beauty spots that are glued on the skin?

Also, those spots are almost black and have basically no colorability. Would it not be better if they are almost white and have that black as default color?
At least I see no justification why they can only be this color?

Jomshir98:
Isn't "Beauty mark" more common name for these?

SandrinePDR:
I renamed and whitened the asset. 
@ClaudiaMia They are meant as part of make-up, so they shouldn't be an integral part of the character's skin

ClaudiaMia (approved):
LGTM!

PR #406, opened 2025-03-24
[ADD] Fingerprint lock

Livie53:
## References

depends on https://github.com/Project-Pandora-Game/pandora/pull/921

## About The Pull Request

Adds fingerprint lock to make use of the added functionality in dependency PR.

## Changelog

Authored by: Livie53
```md
Assets:
- [Fingerprint Lock] Added fingerprint lock
```

## Checklist


src/assets/locks/fingerprint_lock/fingerprint_lock.asset.ts
```diff
@@ -0,0 +1,21 @@
+DefineLockAsset({
+	name: 'Fingerprint Lock',
+	assetPreferenceDefault: 'maybe',
```
Jomshir98, line 3: This lock is even stricter than the combination lock, which is set to 'prevent', so this one should be as well, by default.
```suggestion
	assetPreferenceDefault: 'prevent',
```

Jomshir98 (changes requested):

Jomshir98 (approved):

PR #407, opened 2025-03-25
[ADD] Wings

SandrinePDR:
## References

_None_

## About The Pull Request

Added a foldable pair of wings for demonic and angelic beings.

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [body/wings] New asset
```

## Checklist


src/assets/body/wings/graphics.json
```diff
@@ -0,0 +1,158 @@
...
+{
+	"layers": [
+		{
+			"x": 0,
+			"y": 0,
+			"width": 1000,
+			"height": 1500,
+			"type": "mesh",
+			"name": "angel",
+			"priority": "BELOW_BODY",
```
ClaudiaMia, line 11: I believe, similar to tails, the priorities of all these three layers should be `BELOW_BACK_HAIR` instead, as otherwise most worn assets are visually overlapping them and it looks very weird in the back view.

ClaudiaMia (changes requested):
Really not a fan of having wings in Pandora personally, but they do look nice if someone is into seeing demon or angel characters in a bondage club.

Found one issue.

src/assets/body/wings/wings.asset.ts
```diff
@@ -0,0 +1,78 @@
...
+				{
+					id: 'angel',
+					name: 'Angel',
+				},
+				{
+					id: 'demon',
+					name: 'Demon',
+				},
+			],
+		},
```
Jomshir98, line 42: I again wonder, if the two types share anything in common, or if they would be better split into two assets (same thing as tails).
If they are to be kept as a single asset, the type module should definitely be set as `ADD_REMOVE`.

Jomshir98 (changes requested):

src/assets/body/wings/wings.asset.ts
SandrinePDR: Split the asset into two separate ones

src/assets/body/wings/graphics.json
SandrinePDR: Good point

ClaudiaMia (approved):
LGTM!

Jomshir98 (approved):

PR #408, opened 2025-03-25
[ADD] Bunny accessories

SandrinePDR:
## References

_None_

## About The Pull Request

Added Bunny outfit accessories consisting of a collar with different ties and cuffs

## Changelog


ClaudiaMia (changes requested):
I noticed that you tend to add `requireFreeHandsToUseDefault: true` to a lot of your items. Please note that this is an optional attribute that has the default `false`, when not defined.

I would agree that the wrist watch is `true`, I feel uncertain about the bunny collar, but I feel the bunny cuffs are likely something that one can easily strip off, by rubbing your wrist against some item. So the true does not seem to make really sense to me, personally.

But this was just a general remark. The assets look great and work nice, but I noticed that this PR also contains two school uniform assets that are the content of another PR. Please clean up the PR, so that there is only that inside what is described.

ClaudiaMia (approved):
Great~

PR #415, opened 2025-04-03
[ADD] Pull-up Diaper

SandrinePDR:
## References

_None_

## About The Pull Request

Added a pull-up diaper as a new underwear

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Pull-up diaper] New asset
```

## Checklist


src/assets/panties/pullup_diaper/pullup_diaper.asset.ts
```diff
@@ -0,0 +1,64 @@
...
+			'Underwear',
+			'Underwear_diaper',
+		],
+		requires: [
+			'!Underwear_panties',
+			'!Underwear_pants',
+		],
+	},
+	modules: {
+		wet: {
```
ClaudiaMia, line 26: As per the Discord discussion and the views of Clare, Livie, Ellie, and me, please remove this module from the asset.

The graphics can be too off-putting if a random user is unexpectedly exposed to them in a public space.
I understand that the graphics go a long way for people who enjoy this fetish and that is fully fine and something we want to support. But with the current features Pandora has, it cannot be made so some people have the module and see the graphics and others do not. Thank you.

src/assets/panties/pullup_diaper/pullup_diaper.asset.ts
```diff
@@ -0,0 +1,64 @@
...
+	},
+	// size:240, y:570, centered
+	preview: 'preview.png',
+	attributes: {
+		provides: [
+			'Clothing',
+			'Underwear',
+			'Underwear_diaper',
+		],
+		requires: [
```
ClaudiaMia, line 20: Why did you forbid wearing assets with those two attribute under the diapers? I mean, you surely would not usually, but why outright forbid it? I can definitely imagine the possibility of someone wanting to wear something under them.

ClaudiaMia (changes requested):
I raised two topics, based on the view of Clare and me.

We can talk about it of course, but I do not think this outcome is unreasonable.

src/assets/panties/pullup_diaper/pullup_diaper.asset.ts
SandrinePDR: I removed the module, but kept the graphics there for now, in case we find an acceptable solution in the near future

src/assets/panties/pullup_diaper/pullup_diaper.asset.ts
SandrinePDR: Mainly because of the strap on panties. I couldn't find a better way to avoid that situation

ClaudiaMia: But that is not a problem, is it? I mean, if they are under the diapers, naturally those straps would be seen on the side. That is just as it should be and not an issue? Also you and taja recently added some types where they are not seen like a slip. So I do not see a reason for those requirements personally.

ClaudiaMia (approved):
Good to merge~

PR #418, opened 2025-04-04
[CHANGE] Add pens to nurse dress

SandrinePDR:
## References

#295 Nurse outfit

## About The Pull Request

Added pens as a new option

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Nurse Dress] Pens as additional option
```

## Checklist


ClaudiaMia (commented):
In a very extreme body configuration, one of the pens is floating:
image

A possible solution could be to have an image with only two pens that is swapped in when the waist bone is smaller than -80.

I am not sure that is worth fixing, though, as if a user with such an edge case body shape does not like this result, then they can simply disable the pens.

ClaudiaMia (approved):
Aligned with Sandrine that we will merge the asset as is.

PR #423, opened 2025-04-06
[FIX] Nurse dres pens for small waists

SandrinePDR:
## References

Fixes #418 

## About The Pull Request

Fixed an issue with floating pens for small waist sizes

## Changelog

Authored by: Sandrine

```md
Assets:
- [Nurse dress] Fixed the issue with floating pens for small waists
```

## Checklist


ClaudiaMia (changes requested):
I mean... *giggles* those pens can be added on both sides. So there is symmetry, same as with the character body. So the floating pens are naturally happening on the left as well as the right side. Sorry to be the bringer of bad news~

ClaudiaMia (approved):
Thank you!~

PR #425, opened 2025-04-11
[ADD] Wooden Chair

Shikifet:
## References

_None_

## About The Pull Request

A not so innocent chair where a character can be tied to, with a few accessories to make it more interesting.

## Changelog

Authored by: Shikifet

```md
Assets:
- [Wooden Chair] Added new room device "Wooden Chair" with optional toys and restraints
```

## Checklist


ClaudiaMia (changes requested):
Very very nice asset! I feel something like this was really missing and I am sure it will provide tons fun!

In my tests, I feel it is working very well!

A few comments, though:
I feel like the chair and cushion images could be brightened a bit more without losing shading, so they can be colored in brighter colors, too, if the user so choses.

Also, just for your information, your branch has a merge conflict with the PR I opened earlier, where I renamed `Vulva_protruding` to `Crotch_protruding`. I propose to merge yours first and then I will change this attribute in your asset on my branch.

Lastly, an optional comment, where I am not sure it needs any changes, but I just wanted to provide it as feedback: 
When the slot is empty and some ties are visible, this can become a point of confusion for users why they cannot enter the device. Then again, I feel the chair can really profit from showing the rope ties even on an empty chair for the mood in a room, so I am leaning to say that it is good like it is.

Shikifet:
Thanks for the review! I brightened them a bit more as suggested, to give users more color diversity if they want.

About the merge, its a good approach, nothing to comment there.

And lastly, about showing the ropes even while no character is in the slot, in this case I also like the idea to show them for mood, but being honest, I'm not sure how I could restrict that behaviour (for future assets), but that is something I'll deal in the future.

ClaudiaMia (approved):
Great asset! Thank you!~

PR #429, opened 2025-04-15
[ADD] Sybian

Shikifet:
## References

_None_

## About The Pull Request

A sybian as a new room_device, with lockable options and toys configuration

## Changelog

Authored by: Shikifet

```md
Assets:
- [Shybian] Added new room item "Sybian"
```

## Checklist

- [ ] The change has been tested locally
- [ ] Licensing information and credits were filled in/modified for added/modified assets
- [ ] I understand this change is licensed based on [Pandora's Asset Licensing](https://github.com/Project-Pandora-Game/pandora-assets/blob/master/ASSET_LICENSING.md) information


src/assets/room_devices/sybian/sybian.asset.ts
```diff
@@ -0,0 +1,475 @@
...
+			type: 'typed',
+			name: 'Character Position',
+			staticConfig: { slotName: 'seated' },
+			variants: [
+				{
+					id: 'front',
+					name: 'Front-facing',
+					default: true,
+					properties: {
+						blockSlotsEnterLeave: ['seated'],
```
ClaudiaMia, line 52: No one can enter this asset in any possible state~ 

This line needs to be removed.

src/assets/room_devices/sybian/sybian.asset.ts
```diff
@@ -0,0 +1,475 @@
...
+								},
+							},
+						},
+					},
+				},
+				{
+					id: 'back',
+					name: 'Back-facing',
+					properties: {
+						blockSlotsEnterLeave: ['seated'],
```
ClaudiaMia, line 66: No one can enter this asset in any possible state~ 

This line needs to be removed.

src/assets/room_devices/sybian/sybian.asset.ts
```diff
@@ -0,0 +1,475 @@
...
+		},
+		cover: {
+			name: 'Cover',
+			default: '#D51B1B',
+		},
+		ring: {
+			name: 'Ring',
+			default: '#D0D0D0',
+		},
+		dildo: {
```
ClaudiaMia, line 18: I have a hard time to notice the difference between dildo and no dildo. I really cannot recognize the slightly higher thingie as a dildo being present, personally.

Maybe it would help to make the dildo separately colorable with a darker default color, so it can be recognized as something standing up from the vibrating cover.

Just an idea, not a request for change.

src/assets/room_devices/sybian/sybian.asset.ts
```diff
@@ -0,0 +1,475 @@
...
+		seated: {
+			name: 'Seated',
+			asset: {
+				name: 'Sybian-type Saddle',
+				size: 'huge',
+				poseLimits: {
+					legs: 'kneeling',
+					bones: {
+						character_rotation: 0,
+						leg_l: [[-180, -30]],
```
ClaudiaMia, line 34: -180 seems way too much. If you test that, you see that the character slides down the sybian, essentially.

Unless that is intended, in my tests, double -45 seemed the last reasonable position that looked like one is still sitting on the device. So between -45 on -30 on both sides.

ClaudiaMia (changes requested):
Very nice asset as always!~

But I did find a few things.

src/assets/room_devices/sybian/sybian.asset.ts
Shikifet: Changed

src/assets/room_devices/sybian/sybian.asset.ts
Shikifet: They are usually a single piece, so I think it would be weird if they had different color base

ClaudiaMia: Hmm, true. Then I guess that is how it is.

Shall I merge it then like this?

PR #433, opened 2025-04-18
[CHANGE] Bondage cross improvements

Shikifet:
## References

_None_

## About The Pull Request

Updated the bondage cross to give more personalization options and more play styles:

## Changelog

- Split cuffs attachments between wrists and ankles, so any combination can be done now.
- Split locks into wrists and ankles.
- Moved facing view options to a separate module.
- Moved upside down position to a separate module. It requires having both ankles tied to cross.
- Add optional headrest module.
- Add optional padding module.

Authored by: Shikifet

```md
Assets:
- [Bondage cross] Updated modules of Bondage cross asset for more diversity and play options. Current characters tied to it will be released. We are sorry for the inconvenience
```

## Checklist


src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
```diff
@@ -28,92 +32,167 @@ DefineRoomDeviceAsset({
...
 			staticConfig: { slotName: 'character_slot' },
 			variants: [
 				{
-					id: 'none',
-					name: 'Not attached',
+					id: 'front',
+					name: 'Front-Facing',
 					default: true,
 					properties: {
+						blockSlotsEnterLeave: ['character_slot'],
```
ClaudiaMia, line 49: This line blocks using the asset completely. I remember this problem from some recent PR~
I can only recommend to use a second tab with a second character to control the testing character, instead of using safemode.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
```diff
@@ -28,92 +32,167 @@ DefineRoomDeviceAsset({
...
+								},
+							},
+						},
+					},
+				},
+				{
+					id: 'back',
+					name: 'Back-Facing',
+					properties: {
+						blockSlotsEnterLeave: ['character_slot'],
```
ClaudiaMia, line 63: This line blocks using the asset completely.
I can only recommend to use a second tab with a second character to control the testing character, instead of using safemode.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
```diff
@@ -28,92 +32,167 @@ DefineRoomDeviceAsset({
...
+			type: 'typed',
+			name: 'Wrist',
+			staticConfig: { slotName: 'character_slot' },
+			variants: [
+				{
+					id: 'None',
+					name: 'None',
+					default: true,
+					properties: {
+						blockSlotsEnterLeave: ['character_slot'],
```
ClaudiaMia, line 86: This line blocks using the asset completely.
I can only recommend to use a second tab with a second character to control the testing character, instead of using safemode.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
```diff
@@ -28,92 +32,167 @@ DefineRoomDeviceAsset({
...
 								poseLimits: {
 									bones: {
 										arm_l: -44,
-										arm_r: -44,
 										elbow_l: -21,
+										arm_r: -44,
 										elbow_r: -21,
-										leg_r: -30,
+									},
+								},
```
ClaudiaMia, line 156: Above this line, a `legs: 'standing',` should be added, same as for the other single-sided states. Otherwise, the character was suddenly kneeling in the air in my test.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
```diff
@@ -158,40 +229,115 @@ DefineRoomDeviceAsset({
...
 							},
 						},
 					},
 				},
 			],
 		},
+		rotation: {
+			type: 'typed',
+			name: 'Rotation',
```
ClaudiaMia, line 264: How you made the process with the state flag, I feel it makes things more clear if you rename the module to "Cross rotation".

As I always considered the cross being static and the character being lifted and tied upside down. But with how your headrest rotates automatically, I assume you intend for the cross to rotate while the character is already tied to it by their feet.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
```diff
@@ -28,92 +32,167 @@ DefineRoomDeviceAsset({
...
-				},
 			},
 		},
 	},
 	modules: {
-		chains: {
+		position: {
 			type: 'typed',
-			name: 'Chains',
+			name: 'Facing Position',
```
ClaudiaMia, line 41: Nitpick: "Facing Position" -> "Facing position" for consistency

ClaudiaMia (changes requested):
Those are nice changes and I agree to them. I found a few issues, though.

**Also! I want to raise awareness towards Pandora's contributing guideline.**
The modification policy the cross asset uses contains the policy "Fixes and New uses, otherwise ask".
The guideline ( CONTRIBUTING.md) defines the "new uses" part as "The author allows anyone to make this asset compatible with new features by modifying it or creating new variants".

You made changes to the asset that are clearly not falling under fixes or new uses. Since I was the original author, you creating this PR is akin to asking me, so it is fine. That said, ideally you ask first, before investing work, since the author could want to say "no" and then it is awkward for everyone.

This is especially important if you change the work of another author, where I might assume in good faith that you asked the original author for permission - though I might actually check in such a case possibly.

All good, just wanted to point this out again, as it is important. Loving your ideas and changes towards the cross.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
Shikifet: I thought it as a way to keep the character suspended by both hands, but if you think it could lead to confusion, I can add it.

Shikifet:
You are right, I didn't tell beforehand because I knew you would be the one reviewing it and I would have changed anything you wouldnt comply about withouth any issue.
Said that I'm aware when changing assets of others contributors. For example, I asked taja before fixing one of the hairstyles for the kigurumi mask.

src/assets/room_devices/bondage_cross/bondage_cross.asset.ts
Shikifet: But in the end, its probably better to also force standing in this case as the others, as it can be felt as a bug instead of a feature (pun not intended). So I added it as well

ClaudiaMia (approved):
Thank you~

PR #460, opened 2025-06-28
[ADD] Handcuffs

SandrinePDR:
## References

_None_

## About The Pull Request

Added handcuffs as new arm restraint 

## Changelog

Authored by: Taja & Sandrine

```md
Assets:
- [Handcuffs] New asset
```

## Checklist

### DIFF
--- a/src/assets/arm_restraints/handcuffs/graphics.json
+++ b/src/assets/arm_restraints/handcuffs/graphics.json
@@ -30,6 +30,11 @@
 									"value": 162
 								},
 								{
+									"bone": "elbow_l",
+									"operator": ">=",
+									"value": -16
+								},
+								{
 									"view": "front"
 								}
 							]
@@ -112,7 +117,12 @@
 								{
 									"module": "cuffState",
 									"operator": "=",
-									"value": "above_head"
+									"value": "front"
+								},
+								{
+									"bone": "elbow_l",
+									"operator": "<",
+									"value": -16
 								},
 								{
 									"view": "front"
@@ -150,6 +160,11 @@
 									"bone": "elbow_l",
 									"operator": "<",
 									"value": 162
+								},
+								{
+									"bone": "elbow_l",
+									"operator": ">=",
+									"value": -16
 								},
 								{
 									"view": "front"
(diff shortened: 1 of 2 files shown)


ClaudiaMia (commented):
It's a wonderful asset.

I do have two improvement suggestions, though:

1. It will likely feel unrealistic not only to me that the module states lead to frozen poses, as your hands are blocked and you cannot select another expected to be possible state yourself.
But I believe that with a handcuffs model with a chain in between like this one, you are likely able to still freely switch between the states "in front low", "in front up", and "above head" as one cluster, and "cuffed behind, low", "cuffed behind, right hand up", as well as "cuffed behind, left hand up" as another, with "cuffed behind, up" staying a separate module state like currently.

So this can be achieved by only having three module states that are then frozen for the handcuffed character, and the rest works via pose changes. You can see this approach in action for the leather and steel wrist cuffs, e.g. steel_wrist_cuffs.asset.ts#64-94. Simply by defining the switchable poses as poseLimits options.
When you switch poses, Pandora switches to the nearest allowed pose, so the user does not need to know the exact pose values for this to work dynamically.

2. You or taja likely cut out a part of the handcuffs image, so the arms would still match the gap in the widest body setting. But when the arms are thin, it leads to gaps, as can be seen in the following screenshot.
There is an approach that leads to better visual quality: Not deleting the gap from the original image, but cutting it out into a separate image that is then layered behind the arms. I use this approach for the room devices, for instance to avoid any kind of gaps or overlapping the body where it looks unclean.
image

ClaudiaMia (changes requested):
In the above head state, it seems you might need to fix the arm order to be right first, as otherwise this happens:
image
Same issue here:
image
and here:
image
also here from the front:
image




in the same module state, if you turn around, the other side can look like this in many poses, independent of the arm order:
image


Also, maybe I phrased it misleadingly in my comment before: I wondered if this following pose is maybe the 3rd individual module state, as I am not sure you can switch between this and the tied behind the hips pose freely, but maybe you can.
image
The current 3rd pose over the head is something you can easily switch to by raising your front cuffed hands. There is nothing holding you back from just lifting your arms into that pose. certainly not the cuffs. So maybe there should only be 2 module states. But I guess we can also simply leave the modules like they are right now.

PR #472, opened 2025-08-08
[CHANGE] Suspension Ring

Shikifet:
## References

_None_

## About The Pull Request

- Fix a bug where rock over ankles where not properly drawn when chest line was on suspended high position (it was drawn at knee height instead of ankle height)
- Add new options to rock, supporting kneeling position if character is frogtied with Leg Jute Rope asset

## Changelog

Authored by: Shikifet

```md
Assets:
- [Suspension Ring] Fix glitch on rock tied to ankle and let it be tied on frogtied position
```

## Checklist


ClaudiaMia (approved):
It took me a bit to achieve the new poses, but it seems all works well and I did not spot any issues from my not systematic testing.

One very minor thing I noticed:
I wonder if the rope from the stone should not be layered under the frogtie rope. At least that is my understanding of how it would be tied?

Shikifet:
I also think that thigh rope should be under the frogtie one. But as its drawn from the ring, as a room device it can only draw graphics in front or behind the character, but not between character layers, so it always be shown over the frogtie graphic

ClaudiaMia:
> I also think that thigh rope should be under the frogtie one. But as its drawn from the ring, as a room device it can only draw graphics in front or behind the character, but not between character layers, so it always be shown over the frogtie graphic

Ah yes, I did not remember that it is not part of the asset, but a different one. 

I guess the only option would be to make a hole into the image of the vertical rope. But not sure that would align good enough for all leg sizes. Likely not worth the effort. I will merge this then.

PR #473, opened 2025-08-08
[ADD] Heterochromia

SandrinePDR:
## References

A request on discord

## About The Pull Request

Added the option to have differently colored eyes aka heterochromia

## Changelog

Authored by: Sandrine

```md
Assets:
- [Eyes 3, Eyes 4, Eyes 5] Added heterochromia as an option
```

## Checklist


ClaudiaMia (commented):
The CI warnings seem to come from a left/right definition problem.

Pandora considers the left/right eye from the perspective of the character, but the committed images are named left/right based on how the user looks at it on the computer screen.

While the asset functionally works, the validation rightfully warns that the implementation is a hack and the definition is not fully correct.

For instance for the left iris layer, the left eye point filter was selected, but then it was mirrored so that it works as you desired, since the images are not left from Pandora's perspective. But for Pandora, this mirroring now means that both sides should contain an image, the original side and the mirrored one, but there is only an image on the mirrored side and therefore it screams that there is an empty layer on the original side.

How it actually would be correct is to not mirror the iris left layer and rename all *_right files to _left.

Same issue and solution for the right layer and the according file names respectively.

@Jomshir98 : Can these fixes be done or will you insist in your review that this asset needs to be redone with automatic image layers (auto meshes)?

Jomshir98 (changes requested):
First of all, changing images is not needed at all. In fact doing so is counterproductive and all it will do is increase the size of the repository and makes it actually _harder_ to work with in the assets - so this should be undone.

(I will be away until the end of the week, so rest is written purely from memory)
Currently, mirroring logic is not implemented for `colorKey`, so there are two options for how to do about this, keeping in mind that all layers in question already use mirroring:
- Implement mirroring for `colorKey` in the main repo using `MirrorBoneLike`. After that the only things that are needed in this repo is:
  - Duplicate all colors with `_l` and `_r` suffixes
  - Update color links for all layers
- Alternatively - harder option, but without changes to the main repository:
  - Disable mirroring for all layers, duplicate them and mirror each manually
  - Update the color definitions same as in the first option
  - Update color links same as in the first option

In either case the point type filter is what takes care of only using part of each source texture, so manual splitting is not needed.

Jomshir98:
Project-Pandora-Game/pandora#32 would also be a great thing to have before this in order not to make the current (arguably much more common) usecase - of having the sides colored the same - harder to use.

SandrinePDR:
Closed due to new technical options after last MR

PR #485, opened 2025-09-07
[ADD] Bar Stool

Shikifet:
## References

_None_

## About The Pull Request

A new room device, a bar stool with 2 slots.
 - A sitting slot where the character can have the restraints of their arms attached to the bar stool, preventing her from leaving. It should work with most arm restraints that forces your arms in your back, drawing behind the backrest. Ideal to use with ropes, handcuffs and similar.
 - A kneeling slot below it, with the head in a pillory. Belts can be in two states, for a more casual usage (with other restraints like armbinder) or tightened if the clothing is fitting enough.

## Changelog

Authored by: Shikifet

```md
Assets:
- [Bar Stool] New bondage room device 'Bar Stool'
```

## Checklist


ClaudiaMia (changes requested):
Found one more issue with the order of the belt buckle image layering:

image

Shikifet:
Fixed! Could it be possible I didn't see it due to hair being shown over the belts?

In fact there were 2 issues, as bottom belts were shown correctly because their layer wasn't properly updated to the new ABOVE_BACK_HAIR priority, but all of them are working properly now

PR #490, opened 2025-09-12
[CHANGE] Add new graphics for rope end long while suspended

Shikifet:
## References

_None_

## About The Pull Request

Add graphics for rope end long tie, applied when character rotation is changed, to show a more natural flow of the rope while suspended in both bamboo and ring.
When character is fully upside down, rope has been made shorter to avoid 'trespassing' behind hair

## Changelog

Authored by: Shikifet

```md
Assets:
- [Arm Jute Rope] Updated rope end long graphics shown when character is suspended
```

## Checklist


src/assets/arm_restraints/arm_jute_ropes/graphics.json
```diff
@@ -794,6 +794,23 @@
...
+							[
+								{
+									"module": "arms",
+									"operator": "=",
+									"value": "wrists_high"
+								},
+								{
+									"bone": "character_rotation",
+									"operator": "=",
+									"value": 180
```
ClaudiaMia, line 809: 180 and -180 are the same upside down rotation. So, just 180 is not enough.

But I would recommend to maybe mirror the 26 degree leeway you gave when upright and make the condition similarly:
when bigger than 154 OR when smaller than -154.

ClaudiaMia (changes requested):
Nice!~

Is this PR finished? Just asking, because of that question you asked on Discord about a related topic.

Have one finding.

ClaudiaMia (approved):
Looking good to me now. Thank you!~

PR #626, opened 2026-06-15
[ADD] Anal beads

Shikifet:
## References

_None_

## About The Pull Request

Add a new toy 'Anal beads', with number of beads configurable

## Changelog

Authored by: Shikifet

```md
Assets:
- [Anal beads] Add new toy 'Anal beads'
```

## Checklist


src/assets/toys/anal_beads/anal_beads.asset.ts
```diff
@@ -0,0 +1,146 @@
+DefineAsset({
+	name: 'Anal beads',
```
ClaudiaMia: Just a very minor thing: We capitalize all asset names, so b -> B.

But I could also do that another time.

src/assets/toys/anal_beads/graphics.json
ClaudiaMia, line 1: Wow, that's an impressive definition!

Not sure you feel like doing that this time for this asset, since the code is so complex, but maybe next time?
Clare wants all new JSON files to use automesh layers, as far as possible, as they make certain things much easier for an asset maker and also make such a code much shorter and more efficient. They can be easily made in the asset editor.
I mean, for the room graphics, you were already using the automesh layer definition near the end of the file.

If you have questions on that part, like for your next asset, I am happy to help or point you to some good examples!~

ClaudiaMia (changes requested):
Amazing asset! Works and looks really well!~

Preeetty sweet how you can configure and use it!! Great addition and idea!

Just a minor thing and an (optional) comment, where I could understand that it is too much work.

src/assets/toys/anal_beads/graphics.json
Jomshir98: The most important thing about automesh layers is, that they are much less error prone than old style of meshes, especially as Pandora evolves over time.

Shikifet: Will take that into account, I also found automeshes very useful in the past (for example when creating room graphics of suitcase). In fact I thought in using them for this one, but having a layer per bead, and so many options in modules for each one, I had the initial perception that it would be very hard to manage correctly.
Later when I saw how complicated it was becoming, I was not so sure if it was indeed easier like this, but as you said, at that point I didn't feel like changing idea.

src/assets/toys/anal_beads/anal_beads.asset.ts
Shikifet: Fixed!

ClaudiaMia (approved):
Thank youuu!~

PR #627, opened 2026-06-15
[ADD] Hand pump

SandrinePDR:
## References

_None_

## About The Pull Request

Added two new assets "breast pump". One item that can held in a hand and once to be attached to somebody's breasts.

## Changelog

Authored by: Sandrine & Tom

```md
Assets:
- [Breast pump (handheld)] New handheld item 
- [Breast pump] New item that can be attached to somebody's breasts for milking
```

## Checklist


src/assets/handhelds/handheld_breast_pump/handheld_breast_pump.asset.ts
```diff
@@ -0,0 +1,171 @@
...
+			name: 'Fluid',
+			default: '#FFE599',
+		},
+	},
+	roomDeployment: {
+		autoDeployRelativePosition: [350, -50, 0],
+	},
+	// size:175, y:400, X:655
+	preview: 'preview.png',
+	attributes: {
```
ClaudiaMia, line 42: These attributes should be moved from the top level to the modules, so when the item is used in the left and/or right hand.

Otherwise they interfere with the attached to bust module states in weird ways.

src/assets/handhelds/hand_pump/hand_pump.asset.ts
```diff
@@ -0,0 +1,171 @@
...
+					name: 'Almost full',
+				},
+				{
+					id: 'full',
+					name: 'Filled',
+				},
+			],
+		},
+	},
+	chat: {
```
ClaudiaMia: You can remove those custom chat messages (so the whole chat: {} section) since every little bit helps when the Internet is bad for someone and data rates are low, as the messages are almost exact worded duplicates of the default messages that all items have if no custom ones are defined in the asset definition file here.

ClaudiaMia (changes requested):
Works well! Found just two minor things.

But a general hint: An asset that is both a handheld and a wearable toy(?) is not how we usually split assets and makes it also hard to find something specific in the wardrobe.

We usually make two assets in those cases. Like handheld dildo and dildo toy, or vibe wand and crotch rope vibe wand, or handheld rope and rope.

The reason to keep it separate is that at some point we want to add a feature to transform compatible assets within defined groups when you move them from character to character or from inventory to character.

So you could give the pump to someone else and pick if it stays a handheld or gets a toy. Much less awkward to use than if it stays as one item.

It is fine to not put in this work for now, but eventually it has to be split for consistency.

src/assets/toys/breast_pump/breast_pump.asset.ts
```diff
@@ -0,0 +1,145 @@
...
+					id: 'almost',
+					name: 'Almost full',
+				},
+				{
+					id: 'full',
+					name: 'Filled',
+				},
+			],
+		},
+		fillLevel_l: {
```
ClaudiaMia: You do not need two modules for fill levels for left and right. First of all, this was not working anyway in my tests, and second it is sufficient to have one module like for the handheld asset that then shows the state in all states, none, left, right and on the floor.

A user anyway needs to use two assets to cover both breasts.

ClaudiaMia (changes requested):
Great rework, works good!

Found one issue.

ClaudiaMia (approved):

PR #639, opened 2026-07-04
[CHANGE] Split breast covering attribute into variants

SandrinePDR:
## References

No issue to reference directly, but this change tackles the fact that a breast pump could only be attached to one breast.

## About The Pull Request

Split the attribute `Breast_cover` into two variants for left and added a second attribute pair `breast_protruding` that can be used for bigger breast items.

## Changelog

Authored by: Sandrine

```md
Assets:
- [Breast pump] Can now be used on left and right breast at the same time
- [Chopstick nipple clamps] Added the new protruding attribute
- [See through chastity bra] Cover can not be closed now, if a big item is worn on the breasts
- [Bandeau bra, Style 1 bra] Can not be worn over a big breast item any longer
- [Latex corset] Can not be worn in the breast covering variant over a big breast item
```

## Checklist


src/attributes.ts
```diff
@@ -547,9 +547,21 @@ const ATTRIBUTES_DEFINITION_BASE = DefineAttributes({
...
+	Breast_cover_l: {
+		name: 'An item covering the left breast',
+		description: 'An item that is covering the left breast',
+	},
+	Breast_cover_r: {
+		name: 'An item covering the right breast',
+		description: 'An item that is covering the right breast',
+	},
+	Breast_protruding_r: {
+		name: 'A big item covering the right breast',
```
ClaudiaMia: I would rephrase those similar to the other protruding attributes, so:
'An item significantly protruding from the right breast'

Reason being that the word covering is not accurate enough. "Covering" in Pandora implies that you cannot remove anything from under it, but these attributes does not have such an effect so the word is misleading and could be confusing. It is not guaranteed that the item is always big enough to cover the whole breast.

src/attributes.ts
```diff
@@ -547,9 +547,21 @@ const ATTRIBUTES_DEFINITION_BASE = DefineAttributes({
...
+		description: 'A big item that is covering the right breast',
+	},
+	Breast_protruding_l: {
+		name: 'A big item covering the left breast',
```
ClaudiaMia: I would rephrase those similar to the other protruding attributes, so:
'An item significantly protruding from the left breast'

src/attributes.ts
```diff
@@ -547,9 +547,21 @@ const ATTRIBUTES_DEFINITION_BASE = DefineAttributes({
...
+		name: 'An item covering the left breast',
+		description: 'An item that is covering the left breast',
+	},
+	Breast_cover_r: {
+		name: 'An item covering the right breast',
+		description: 'An item that is covering the right breast',
+	},
+	Breast_protruding_r: {
+		name: 'A big item covering the right breast',
+		description: 'A big item that is covering the right breast',
```
ClaudiaMia: 'An item that is significantly protruding from the right breast'

src/attributes.ts
```diff
@@ -547,9 +547,21 @@ const ATTRIBUTES_DEFINITION_BASE = DefineAttributes({
...
+	},
+	Breast_protruding_l: {
+		name: 'A big item covering the left breast',
+		description: 'A big item that is covering the left breast',
```
ClaudiaMia: 'An item that is significantly protruding from the left breast'

src/assets/bras/classic_corset/classic_corset.asset.ts
```diff
@@ -23,6 +23,10 @@ DefineAsset({
 			'Underwear',
 			'Underwear_corset',
 		],
+		requires: [
```
ClaudiaMia: This is incorrect and can be removed. The classic corset does not cover the breasts and hence anything protruding from them does not matter.

src/assets/bras/bandeau_bra/bandeau_bra.asset.ts
```diff
@@ -20,6 +20,10 @@ DefineAsset({
 			'Underwear',
 			'Underwear_bra',
 		],
+		requires: [
```
ClaudiaMia: Similar to how you did it for the latex corset, those attributes should not be added globally but only to the module state 'covered' of the bustState module.

ClaudiaMia (changes requested):
Great, thank you! Have some minor findings.

ClaudiaMia:
Clare said she will fix the CI with the asset hash, when it is ready to be merged.

ClaudiaMia (approved):
Thank you!~

Jomshir98 (approved):

PR #640, opened 2026-07-13
[FIX] Cleaned up low alpha pixels from body_base

Nikky90506:
## References

I don't think there's any particular issue created for this:

picture

### Goal

Turns out all of the layers for Base Body have "stray almost-transparent pixels at random places". The goal was to remove most glaring cases of these.

### How

The idea was to find minimal alpha value threshold at which selecting outer transparent pixels didn't result in any "islands" and then deleting all the pixels in the selection. So threshold was found manually for each layer, I think it was 30 at most. Honestly, it could probably be even more, but I decided to err on the side of caution.

### Example

<details>
<summary>Body_back changes</summary>

__Before:__
Body_back_before

__After:__
Body_back_after
</details>

### Notes

By nature of this change it's very unlikely to be able to see the differences visually, but you may notice the reduced filesize of layers.

It can be checked manually though, by using your preferred editor. For Krita, it would be choosing "Similar Color Selection Tool", setting Tool Options -> Selection extent -> Threshold to minimum and clicking on outer transparent area of the layer.

<details>
<summary>Krita interface</summary>
  Krita

</details>

I didn't like trying to use Krita for the cleanup though, as it handles such selections weirdly and used GIMP instead. But it's enough to see the issue.

It might be a good idea to add this to asset creation tutorial as we have other assets with similar problem. I didn't want to go through the whole repo at once and decided to keep the scope to Body Base. But, for example, Asian Dress has a little of these stray pixels too. Probably some kind of automated check could be devised for this, but that's an idea for another day.

### GIMP vs Krita
<details>
<summary>
Example of differences between Krita and GIMP. While GIMP selects all pixels below Threshold, Krita... I'm not exactly sure what happens.
</summary>
  
  
  __GIMP__
  GIMP-example
  
  __Krita__
  Krita-example

</details>

## Changelog

Authored by: Nikky

```md
Assets:
- [Base Body] Cleaned up stray low alpha pixels from all layers
```

## Checklist

- [ ] Licensing information and credits were filled in/modified for added/modified assets


Nikky90506:
Oh, so tests check hash of core assets...
Though I can't think of a way it could be updated at the same time anyway, considering hash is in another repo.

ClaudiaMia:
> Oh, so tests check hash of core assets... Though I can't think of a way it could be updated at the same time anyway, considering hash is in another repo.

Yes, Jomshir will take care of that.

Thank you for this cleanup!
We will take a look at it soon.

Jomshir98 (approved):

PR #644, opened 2026-07-22
[ADD] Vacbed

Shikifet:
## References

_None_

## About The Pull Request

Vacbed created as a Room device. It has 2 modes (transparent and opaque). Opaque is meant to be used mostly naked, as it outlines body, but can't outline toys or devices over character, while transparent mode can be used more freely.

## Changelog

Authored by: Shikifet

```md
Assets:
- [Vacbed] Added new room device "Vacbed"
```

## Checklist


src/assets/room_devices/vacbed/vacbed.asset.ts
```diff
@@ -0,0 +1,310 @@
...
+						slotProperties: {
+							inside: {
+								attributes: {
+									provides: [
+										'Restraint',
+										'Restraint_mouth',
+										'Mouth_item',
+										'Mouth_insert',
+									],
+									requires: [
```
ClaudiaMia, line 220: I think this also requires

			'!Mouth_tongue_out',
			'!Mouth_protruding',

src/assets/room_devices/vacbed/vacbed.asset.ts
```diff
@@ -0,0 +1,310 @@
...
+					name: 'Outside',
+					default: true,
+				},
+				{
+					id: 'encased',
+					name: 'Inside',
+					properties: {
+						slotProperties: {
+							inside: {
+								attributes: {
```
ClaudiaMia, line 209: I believe this should also cover something:

		covers: [
			'Mouth_item',
		],

src/assets/room_devices/vacbed/vacbed.asset.ts
```diff
@@ -0,0 +1,310 @@
...
+									provides: [
```
ClaudiaMia, line 213: This also provides mouth cover for something like a stuffing, so it cannot be removed while this is applied:

'Mouth_cover',

src/assets/room_devices/vacbed/vacbed.asset.ts
```diff
@@ -0,0 +1,310 @@
...
+						},
+						legs: {
+							pose: 'standing',
+						},
+						options: [
+							{
+								bones: {
+									character_rotation: [[0, 0], [180, 180]],
+									leg_l: [[-18, 22]],
+									leg_r: [[-18, 22]],
```
ClaudiaMia: rotation and leg l/r seems to always be the same for all options.

Therefore you can remove them from all options and just write it once in the section where you also defined the view, so somewhere in lines 38 to 44.

This makes the asset definition code smaller.

ClaudiaMia (changes requested):
Wow! Amazing asset! Impressive pose options, that must have taken ages to figure out!

I suppose now we need a new effect that can be applied when the device is under vacuum that prevents posing. But Clare said this is not so easy to make. Maybe in the future. It does not make this asset less great to use for now!

Is it intentional that there is no lock slot that can simulate to prevent someone from letting out the air from the air entry plug? Just a curious question, as we use lock slots quite liberally.

Found a few small things.

ClaudiaMia:
Clare will fix the failing CI when this is ready and approved.

Shikifet:
> Wow! Amazing asset! Impressive pose options, that must have taken ages to figure out!
> 
> I suppose now we need a new effect that can be applied when the device is under vacuum that prevents posing. But Clare said this is not so easy to make. Maybe in the future. It does not make this asset less great to use for now!
> 
> Is it intentional that there is no lock slot that can simulate to prevent someone from letting out the air from the air entry plug? Just a curious question, as we use lock slots quite liberally.
> 
> Found a few small things.

Thank you! I fixed all comments.

I don't mind if there is no pose limit option. Indeed, I enjoy seeing the character moving with limited options, as it seems is really struggling against the latex.

And about lock, I was so focused on making the layers working as I intended, I totally forgot about it. But there is no way to fully prevent anyone from escaping without one, so of curse I added it. I also moved the status module to bottom, as I usually like leaving the device settings on top, and interactive ones below.

One last thing, I tested it always against a wall, but now I realized that maybe it should include something to hang it on the ceiling (like the chains on Bondage Cross). What do you think?

Hoping everybody having fun with it!

ClaudiaMia:
Thank you!

> One last thing, I tested it always against a wall, but now I realized that maybe it should include something to hang it on the ceiling (like the chains on Bondage Cross). What do you think?

That sounds like a great idea as an option!~

> Hoping everybody having fun with it!

Oh, I am sure it will be a very fun asset that will be enjoyed by many!!!

ClaudiaMia (approved):
Amaaaaaaaaaaaaaaaaaaazing asset!! Thank you so much for this!~

PR #645, opened 2026-07-23
[ADD] Shapes

DoppelChest1:
## References

_None_

## About The Pull Request

Added a new modular room device "Shapes", that is a set of simple geometric shapes (square, triangles, circle, diamond, pentagon, hexagon, star, square with round/square hole, ring and horizontal/vertical bars) that players
can combine on a single item to build custom decorations.

Each is made as an independent module, so multiple shapes can be enabled at once and combined together. Every shape supports four sizes (Small, Medium, Big, Huge)

There is also an additional Text module, that also will allow for players to use text without another object visible, since all shapes can be disabled.

## Changelog

Authored by: DoppelChest

```md
Assets:
- [Shapes] - Added new room device "Shapes" - a set of combinable geometric
  shapes for building custom room decorations
```

## Checklist


ClaudiaMia (approved):
Great work! This will be very useful for room designing! Thanks a lot!

PR #649, opened 2026-07-30
[ADD] Added Belted, Capri, and Bermuda variants to Jeans

freyja-veritas:
## References

_None_

## About The Pull Request

Added more customization options to the existing Jeans asset:

- Added an optional belt with front and back graphics.
- Added Capri and Bermuda styles with matching front and back graphics.
- Added folded cuff graphics for the Capri and Bermuda styles.
- Added Freyja to the asset credits for the new variants.
- Standardized the fabric reverse-side color for Jeans and Jeans Shorts to `#c2cfda`.

This gives players shorter jeans silhouettes and an optional belt without requiring separate assets.

## Changelog

Authored by: Freyja

```md
Assets:
- [Jeans] Added Belted, Capri, and Bermuda variants.
```

## Validation

- `pnpm exec eslint src/assets/bottoms/jeans/jeans.asset.ts`
- `pnpm run type-check:src`
- `pnpm run build`

`pnpm run build --prod` was also attempted. The local Git 2.24.1 installation does not support the `git shortlog --group` option used by the build's contributor check, so production mode treats that environment warning as fatal. The non-production asset build completed successfully.

## Checklist


src/assets/bottoms/jeans/jeans.asset.ts
```diff
@@ -8,10 +8,18 @@ DefineAsset({
 			name: 'Pants',
 			default: '#9BCDFF',
 		},
+		inside: {
+			name: 'Fabric reverse side',
+			default: '#c2cfdaff',
```
ClaudiaMia: This should just be `#c2cfda``. The "ff" denotes an alpha value, but transparency is not set for this color group.

Having the "ff" here leads to an issue when using the color reset button.

And by the way, same issue with the "jeans shorts" asset where you did the same. Also here the "ff" needs to be removed for this color. I did not spot that back then, so would be great if you could fix that also in this pull request.

ClaudiaMia (changes requested):
Great extensions, love them!
Found just one issue.

Also, please feel free to put your your name to the "credits" of the asset, as we have done for the "jeans shorts".

src/assets/bottoms/jeans/jeans.asset.ts
freyja-veritas: done and done ❤️

ClaudiaMia (approved):
Thank you!

PR #652, opened 2026-08-03
[ADD] Added new asset Butterfly pendant

smallelise:
## About The Pull Request

Added a new Collar Accessory asset.
A butterfly that attaches to the ring on all existing collars with a ring module.

## Changelog

Authored by: smallelise

```md
Assets:
- [Butterfly pendant] A butterfly that attaches to the ring on all existing collars with a ring present.
```

## Feedback to asset before Pull 

Below in the image is the attachment visible. I need feedback on the chain at the top. 
Is it okay like it is, or should it be thicker and more visible?

image

## Checklist


src/assets/accessories/butterfly_pendant/butterfly_pendant.asset.ts
```diff
@@ -0,0 +1,70 @@
...
+		details: {
+			name: 'Details',
+			default: '#FFE100',
+		},
+		wings: {
+			name: 'Wings',
+			default: '#090B0F',
+		},
+		wing_fabric: {
+			name: 'Wing Infill',
```
ClaudiaMia, line 22: Could be an optional idea to allow some transparency for this color and maybe the "details" one, by adding ,e.g. `minAlpha: 0.5,`.

Wings can be a bit see-through in real life.

src/assets/accessories/butterfly_pendant/butterfly_pendant.asset.ts
```diff
@@ -0,0 +1,70 @@
...
+			default: '#DBA62B',
+		},
+		chain: {
+			name: 'Chain Link',
+			default: '#A1A1A1',
+		},
+	},
+
+	//ToDo: roomDeployment: {autoDeployRelativePosition: [0, -10, 0]},
```
ClaudiaMia: Please either remove this TODO comment or add the floor graphics.

src/assets/accessories/butterfly_pendant/butterfly_pendant.asset.ts
```diff
@@ -0,0 +1,70 @@
...
+		credits:
+			['smallelise'],
+		modificationPolicy:
+			'Fixes only, otherwise ask',
+		reusePolicy:
+			'Free to use',
+		licensing:
+			[
+				{
+					part: 'Shark-Plush',
```
ClaudiaMia: Sharkie!~

src/assets/accessories/butterfly_pendant/graphics.json
```diff
@@ -0,0 +1,47 @@
+// THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+{
+	"layers": [
+		{
+			"x": 375,
+			"y": 285,
+			"width": 250,
```
ClaudiaMia: The images are 1000x1000 and you use them as 250x250, meaning 4x the resolution.

We use this approach in a few assets, like steel cuffs, where it is 1.5x and one set of eyes uses 2x.

4x is a bit extreme.

Please resize the images to just 500x500, so the resolution is only doubled.

ClaudiaMia (changes requested):
Lovely new asset!~ 

Thank you!

I have a few comments.

ClaudiaMia:
>  I need feedback on the chain at the top.
Is it okay like it is, or should it be thicker and more visible?

I feel it is fully fine like this!~

src/assets/accessories/butterfly_pendant/butterfly_pendant.asset.ts
smallelise: Added! Actually, a good idea.

src/assets/accessories/butterfly_pendant/butterfly_pendant.asset.ts
smallelise: Hihi, gocha, another Sharkie.. >.<

src/assets/accessories/butterfly_pendant/graphics.json
smallelise: That's not correct 99% of The Image is empty, just to satisfy the Compiler requirements and avoid artifacts.
The actual Asset size is: ~128 x 128px.

~~width~~ != Asset Size!

image

Jomshir98: But graphics cards treat empty and used pixels the same. And Pandora can only optimize images on the edge of mesh points cut (which is why it requires this size of image), as it cannot recalculate transforms for images that don't cover the used area of mesh.
And GPU needs to use raw, uncompressed texture, which means 4 bytes per pixel, no matter if the pixel is empty or not. Which means that VRAM usage at runtime for this one image is:
image

ClaudiaMia: So please resize all images to 500x500 to reduce the otherwise monstrous VRAM usage of this little asset, since it consists of five such images.

smallelise: I will try to resize the image.
I'm currently unsure if that will bring back the waring with unused space in all layers because the asset is directly over one cross of such an mesh intersection.

Scaling the asset further down below 128px actually size as shown above would introduce a new problem from my perspective that it would be impossible to make out any details.

Regardless i will create a 500px version. and pose the result here for future discussion.

ClaudiaMia: Hm... I see what you mean. 

I will discuss again with Clare this evening.

smallelise: Might it be possible and maybe improve even performance to cut the existing meshes in half? so one sector becomes 4 individual ones.


Then the asset itself can easily be made smaller in pixel size and at the same time keep resolution.

ClaudiaMia: > Might it be possible and maybe improve even performance to cut the existing meshes in half? so one sector becomes 4 individual ones.
> 
> Then the asset itself can easily be made smaller in pixel size and at the same time keep resolution.

Clare has already experimented with this.
But while more points means finer control and such better optimization and lower VRAM usage, it also increases computational complexity for each draw, which is directly proportional to triangle count, which is a function of a square of how many cuts there are in static template.
The current template has this many cuts, because that originally put it at roughly the same complexity that the body template had back then.

And about the original topic:
I think we both misunderstood how it works.

I asked Clare again and she said:

There is currently the optimization that only active part of the mesh that overlaps with some visible pixels defines a rectangle that gets cut out~ Limitation of current implementation just is, that it works like this:
- Get list of triangles that should be used based on layer settings
- Figure out which of these triangles have at least one pixel that isn't fully transparent inside
- Find the smallest rectangle that contains all triangles selected in the previous step
- And use that rectangle to cut a part of the original texture which gets actually loaded

So basically, if your layer has the size of 250x250 then it is a big difference if the image is 1000x1000 or 500x500 as those later cut meshes contain a higher pixel density since you use 4x resolution, increasing VRAM usage for weaker devices.

Normally we define the layer as 1000x1500 and the image as 1000x1500 and no matter how small or big the asset in the image is, it is optimally cut at same resolution.

So your 4x times resolution approach is too expensive. 2x is borderline okay, as I understand you want some increased sharpness for the pendant.

Another thing to consider is that there isn't that much value in it having too much higher resolution to other assets, because by the time you zoom in enough to see the better resolution, everything else will be a blurry mess.

smallelise: Changed the resolution to 2000x3000 and adjusted the width accordingly.

Does it make a difference? See below:
500x500 with a width of 250px OR 2000x3000 with a  width and height of 2000x3000?

I need to add that I tried 1x Resolution, and then it is just a blurry mess all the time.

Jomshir98: > Does it make a difference? See below: 500x500 with a width of 250px OR 2000x3000 with a width and height of 2000x3000?

For users 500x500 in 250x250 window vs 2000x3000 in 1000x1500 (I assume you wrote typo) is the exact same if all the additional pixels are empty - Pandora's tooling with optimize them away during build.
For repository itself 500x500 is slightly preferred, as the image file (.png) will come out smaller - this doesn't impact users at all, but will help developers in the long run, as when you clone a repository you have to download all the files in it, which adds up with many assets.

What matters is 1000x1000 in 250x250 layer vs 500x500 in 250x250 layer.
The "layer" size actually doesn't matter for resource usage at all. You could make that layer 100000x100000 and it would use the exact same resources - layer size only describes how the image is rendered. So bigger layer size would stretch the image, smaller layer size compresses it (i.e. increases quality).
What matters for performance is count of pixels of the texture, which is a 4x difference in this example.

> I need to add that I tried 1x Resolution, and then it is just a blurry mess all the time.

Yes, I agree up to 2x resolution makes sense for small, detailed assets. Claudia already mentioned existing example use. But more than 2x doesn't fit with Pandora's overall resolution targets we decided to use - it uses substantially more resources without much benefit, because everything else looks blurry by comparison, making such zoom unusable.

smallelise: Then i keep it at 2x and will try to make the image size smaller while keeping the ratio the same.

This matter is so far resolved then?!

ClaudiaMia: Yes, but topic is only resolved, when we re-review the pushed changes and it is confirmed to be all good now from our side as well.

smallelise: _Should_ be obvious to all yes >.>
Some voices elsewhere saying that's not so obvious... on similar topics.

ClaudiaMia: I am not sure I understood you correctly.

I just observed that you resolved some other comment in this PR without committing new code that fixes it, therefore I felt like pointing it out here, just in case.

Also ideally, the one who adds the review comment should be the one to resolve it.

In any case, no harm done and please take any time you need~

smallelise: Okay next timw i will let te one who committed resolve it. thought it makes my organization better.

the changes are as of now only locally >.> didn't push anything.

src/assets/accessories/butterfly_pendant/graphics.json
```diff
@@ -0,0 +1,89 @@
...
+					"body.png"
+				]
+			}
+		}
+	],
+	"roomLayers": [
+		{
+			"x": 0,
+			"y": 0,
+			"width": 1,
```
ClaudiaMia: Shouldn't those be 400 width and 400 height, as otherwise the pendant does not show on the floor.

Also, x and y should likely not be "0" as then the item on the floor is quite a bit away from the item button.

ClaudiaMia (changes requested):
Previous things are resolved, but found one more issue with the floor graphics.

src/assets/accessories/butterfly_pendant/graphics.json
smallelise: Ah dame in the previous commit it's correct.... I'll change it again.
_Is sloppiness contagious?_

ClaudiaMia (approved):
Love it! Thank you!~

PR #656, opened 2026-08-13
[ADD] Added new fishnet sleeves asset

Nikky90506:
## References

_None_

## About The Pull Request

Added a new "gloves" asset.

Not sure about attributes. I think existing ones for gloves in it's current state doesn't bring enough to justify it shortcomings.
But of course will add whatever you think is necessary.

## Changelog

Authored by: Nikky

```md
Assets:
- [Fishnet Sleeves] Added new "gloves" 'Fishnet Sleeves'
```

## Checklist


ClaudiaMia (changes requested):
Love it! Such a great fit with the fishnets.

But yes, about the attributes:

Without any you cannot find the asset in any of the categories, other than someone searching it directly or looking for all.
I think so far every asset is matched to one of the (button) categories in the wardrobe.

`Accessory` could be an option, but better might be that it provides `Gloves`, so it appears in the Gloves category. Seems like the best fit to me. What do you think?

Nikky90506:
> Love it! Such a great fit with the fishnets.

Thanks!

> But yes, about the attributes:
> 
> Without any you cannot find the asset in any of the categories, other than someone searching it directly or looking for all. I think so far every asset is matched to one of the (button) categories in the wardrobe.
> 
> `Accessory` could be an option, but better might be that it provides `Gloves`, so it appears in the Gloves category. Seems like the best fit to me. What do you think?

Ah, I was looking at gloves and thinking about cover attributes, didn't realize it needs at least one for category filtering. Thanks, glove it is then.

ClaudiaMia (approved):
Greeeat! Thank youuu!!!~

